Issue #289, opened 2014-11-13
ClickjackFilter after doFilter


meg23:
_From [ronald.p...@googlemail.com](https://code.google.com/u/116937351057973843388/) on August 17, 2012 02:41:28_

I have used the ClickjackFilter but I noticed that in some cases the header simply was not set on the response. I looked at the code and noticed that header is added after chain.doFilter(request, response).

This seems to be a problem at least in Tomcat 6.  I moved res.addHeader("X-FRAME-OPTIONS", mode ) before chain.doFilter(request, response) and it worked fine in every case. 

There is also discussion about this on stack overflow: http://stackoverflow.com/questions/11371755/clickjacking-filter-to-add-x-frame-options-in-response I would suggest moving it before doFilter. 

Best, 
Ronald

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=283_

chrisisbeef:
JIRA Issue ESAPILEG-288

kwwall:
This appears to have been fixed first by commit 2dbb419 and then later again by commit a5c5d3c but it never got automatically closed because the commit comment didn't include the magic verbiage. However, a manual inspection of the doFilter() method shows it to be fixed. It should have been closed years ago, but looks like I missed it last release.

PR #343, opened 2015-01-01
Fix for issue #289

arpitgupta14:
https://github.com/ESAPI/esapi-java-legacy/issues/289

When using ClickJackFilter, In some cases response header is not set.
http://stackoverflow.com/questions/11371755/clickjacking-filter-to-add-x-frame-options-in-response


chrisisbeef:
Looks great - thanks for the fix!

Issue #344, opened 2015-04-02
JUnit test failure in ValidatorTest.testGetValidSafeHTML()


kwwall:
When all the JUnit tests are run against the latest official ESAPI release, there is following failure:

> Failed tests:   testGetValidSafeHTML(org.owasp.esapi.reference.ValidatorTest): expected:<Test.> but was:<Test. alert(document.cookie)>
> 
> Tests run: 540, Failures: 1, Errors: 0, Skipped: 0

The error seems to be coming from src/test/java/org/owasp/esapi/reference/ValidatorTest.java, in testGetValidSafeHTML(). Haven't looked at it any deeper than this.

demkada:
This bug has already been fixed.
Waiting for a pull request validation from

xeno6696:
I think this can be closed, @kwwall

kwwall:
Agree, but I am having some git merge conflicts that need resolved.

Someone just gave me some steps to try today. I will try to give
those steps a shot Tues or Wed.

-kevin

On Mon, Dec 21, 2015 at 4:05 PM, xeno6696 <notifications@github.com> wrote:

> I think this can be closed, @kwwall <https://github.com/kwwall>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ESAPI/esapi-java-legacy/issues/344#issuecomment-166420590>
> .
>



-- 
Blog: http://off-the-wall-security.blogspot.com/
NSA: All your crypto bit are belong to us.

kwwall:
Fixed with commit 89815df046baa67b0f9d3deef22f2c06933dc653.

PR #348, opened 2015-04-27
Fix some issues

demkada:
Fix for issue #344 	948592a
Using the right AccessControlException class to throw Exception 	db10603

### DIFF
--- a/src/test/java/org/owasp/esapi/reference/ValidatorTest.java
+++ b/src/test/java/org/owasp/esapi/reference/ValidatorTest.java
@@ -297,7 +297,7 @@ public class ValidatorTest extends TestCase {
 
         assertEquals("Test. &lt;<div>load=alert()</div>", rule.getSafe("test", "Test. <<div on<script></script>load=alert()"));
         assertEquals("Test. <div>b</div>", rule.getSafe("test", "Test. <div style={xss:expression(xss)}>b</div>"));
-        assertEquals("Test.", rule.getSafe("test", "Test. <script>alert(document.cookie)</script>"));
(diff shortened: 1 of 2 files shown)


src/test/java/org/owasp/esapi/reference/ValidatorTest.java
```diff
@@ -297,7 +297,7 @@ public void testGetValidSafeHTML() throws Exception {
 
         assertEquals("Test. &lt;<div>load=alert()</div>", rule.getSafe("test", "Test. <<div on<script></script>load=alert()"));
         assertEquals("Test. <div>b</div>", rule.getSafe("test", "Test. <div style={xss:expression(xss)}>b</div>"));
-        assertEquals("Test.", rule.getSafe("test", "Test. <s%00cript>alert(document.cookie)</script>"));
+        assertEquals("Test.", rule.getSafe("test", "Test. <script>alert(document.cookie)</script>"));
```
chrisisbeef: Why did you drop the null-byte test here? Like the tests under it, this is verifying that the validation works even when a special character interrupts the payload. The null-byte attack is a valid payload to verify.

demkada: It was a mistake. I wanted to change that line with 
assertEquals("Test. alert(document.cookie)", rule.getSafe("test", "Test. <s%00cript>alert(document.cookie)</script>")); for the test coherency to fix issue #344

demkada: Hello,
Sorry for this absurdity. 
Is it normal that the assertEquals expected object does not correspond to the actual safe object return by the HTMLValidationRule class?

I think the expected object should be equal to "Test. alert(document.cookie)". What is your opinion? maybe I do not dig where I should.

kwwall:
Closed via merge.

PR #357, opened 2016-01-20
Fix issue 355 by adding a call to file.deleteOnExit().

artfullyContrived:
Fixes issue 355 by adding tempFile.deleteOnExit() in InterceptingServletOutputStream.
This only works for a clean vm exit.

If robustness is required I can redo it following the implementation  [here](http://stackoverflow.com/questions/16691437/when-are-java-temporary-files-deleted/16694493#16694493)

### DIFF
--- a/src/main/java/org/owasp/esapi/waf/internal/InterceptingServletOutputStream.java
+++ b/src/main/java/org/owasp/esapi/waf/internal/InterceptingServletOutputStream.java
@@ -56,7 +56,10 @@ public class InterceptingServletOutputStream extends ServletOutputStream {
 		 * the prefix and suffix small for less processing. The "oew" is intended
 		 * to stand for "OWASP ESAPI WAF" and the "hop" for HTTP output.
 		 */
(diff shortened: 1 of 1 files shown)


xeno6696:
@artfullyContrived thanks for your contribution!  @kwwall , the link he nabbed discusses using a trick specific to unix systems, where when you delete a file that is still referenced by a process, it will keep it on the system until the last process terminates it.  

Personally, while that method would certainly work, and most java programs are going to be run on *nix systems, I think relying on the JVM to do its work is a little cleaner, and relies less on *nix implementations.  I recommend taking this change as is.

kwwall:
I think this is good enough for now. It's mostly an annoyance when the JUnit tests run. I'm a firm believe of not letting the perfect become the enemy of the good so I will merge your pull request to close this issue.

artfullyContrived:
Thanks guys. That was quick from you.

kwwall:
Yeah, I shoulda give issue 355 the PITA label I created because each time o
ran 'mvn test' I would leave behind about 40 or so /tmp/oew*.hop files that
I would have to manually remove. So that was motivation. :)

Thank again for the patch.

-kevin

PR #367, opened 2016-04-07
Adding IntelliJ Tests setup documentation

drmyersii:


kwwall:
@drmyersii Thanks David. I'm going to give this a try because I noted it as a vim plugin! :+1:

drmyersii:
@kwwall I'm not sure about the `.codenvy` dirs. I haven't had any show up on my machine, but if IntelliJ creates them, I would definitely ignore those too. Thanks!

PR #370, opened 2016-05-02
Integer overflow in for loop

kravietz:
Fix a theoretical CWE-190 integer overflow as out.length is a long ([RandomAccessFile](https://docs.oracle.com/javase/8/docs/api/index.html?java/io/RandomAccessFile.html)) while the for loop is indexed by an int

### DIFF
--- a/src/main/java/org/owasp/esapi/crypto/CipherText.java
+++ b/src/main/java/org/owasp/esapi/crypto/CipherText.java
@@ -668,7 +668,7 @@ public final class CipherText implements Serializable {
      */
     public byte[] getSeparateMAC() {
         if ( separate_mac_ == null ) {
-            return null;
+            return new byte[0];
         }
         byte[] copy = new byte[ separate_mac_.length ];
         System.arraycopy(separate_mac_, 0, copy, 0, separate_mac_.length);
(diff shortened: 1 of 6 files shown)


xeno6696:
@kwwall we're WAAAAY behind on this issue.  But I don't believe in closing issues if both co-leads don't agree.  I'm willing to accept all 5 commits.  

AFTER my pull request, of course... :-)

xeno6696:
Leave it to me to not think that through.... yeah you're right, the suggested changes simply push the overflow issue to the future.  I forgot my days at Mutual of Omaha and using `BigDecimal`

kwwall:
@xeno6696 Don't beat yourself up for missing that. I've been doing secure code reviews daily as part of my major job function for the past 4 years so I probably am more attuned to that sort of thing than you are. Plus, since it is ultimately coming from a trusted ESAPI.properties file, even the original integer overflow is probably meaningless to begin with. It really was just an attempt to shut up Coverity from complaining.

xeno6696:
Given that the requester has not responded and the fact that the integer in question is NOT controlled by an attacker, I'm going to close this pull request.

Issue #302, opened 2014-11-13
HTMLEntityCodec#decode incorrectly decodes upper-case accented letters as their lower-case counterparts


meg23:
_From [bja...@twigkit.com](https://code.google.com/u/109723178327690826366/) on March 25, 2013 12:05:24_

What steps will reproduce the problem? 1. System.out.println(new org.owasp.esapi.codecs.HTMLEntityCodec().decode("&Aacute;")); What is the expected output? What do you see instead? I would expect the HTMLEntityCodec to correctly decode "&Aacute;" as a capital "Á". Instead, it outputs a lower-case "á". The same is true for all HTML entities whose encoding fits the "&*acute;" pattern. What version of the product are you using? On what operating system? Version 2.0.1 on MacOS X 10.8.2. Does this issue affect only a specified browser or set of browsers? Nope, this is an API issue. Please provide any additional information below. Checking out your source code from trunk (25/3/2013), it seems the problem is line 253 of HTMLEntityCodec.java (in method getNamedEntity): 

```
possible.append(Character.toLowerCase(input.next())); 
```

Here it is turning everything into lower case as it reads the input stream, thereby losing the case information for accented letters.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=296_

sunnypav:
The issue is still happening even with the latest release of ESAPI 2.1.0.1. Please fix this.

xeno6696:
In your debugging you've already identified the line of code.  Why not just fix it and submit a pull request?

sunnypav:
Hi All, Please review the change & let me know the feedback for the same.

sunnypav:
My Apologies for the duplicate pull requests please pull only the commit [005cf26](https://github.com/ESAPI/esapi-java-legacy/commit/005cf26f5aa34b702a243f7b222eb5d2db296543).

madness-inc:
Hi  @xeno6696 ,

are there any plans to release a bugfix version (2.1.0.2) including the fix?
This is a serious bug for us, since german umlauts are also affected.

Regards
Matthias

PR #378, opened 2016-06-08
#302 HTMLEntityCodec Now decodes cased accented letters properly

sunnypav:
HTMLEntityCodec.decode incorrectly decodes upper-case accented letters as their lower-case counterparts. In this change i am doing an exact match for the possible string and if the match is not found falling back to the old technique of lowercasing the possible string & find a match for it. Please review this change

### DIFF
--- a/src/main/java/org/owasp/esapi/codecs/HTMLEntityCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/HTMLEntityCodec.java
@@ -250,12 +250,21 @@ public class HTMLEntityCodec extends Codec
 		// kludge around PushbackString....
 		len = Math.min(input.remainder().length(), entityToCharacterTrie.getMaxKeyLength());
 		for(int i=0;i<len;i++)
-			possible.append(Character.toLowerCase(input.next()));
+			possible.append(input.next());
 
 		// look up the longest match
 		entry = entityToCharacterTrie.getLongestMatch(possible);
-		if(entry == null)
-			return null;	// no match, caller will reset input
+		if(entry == null) {
+			// We are lowercasing & comparing the result because of this all the upper case named entities are getting converted lowercase named entities.
+			// check is there any exact match https://github.com/ESAPI/esapi-java-legacy/issues/302
+			String possibleString = possible.toString();
+			String possibleStringLowerCase = possibleString.toLowerCase();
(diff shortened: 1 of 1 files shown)


xeno6696:
So my first question is this:  What difference does the case make?  Browsers interpret &gt; and &Gt; or &GT; the same:   you get the character `>` rendered.  

What's the use case that triggered the need for this change?  Did you write a unit test for it?  

(If not, I can do it, just give me an example input.)

xeno6696:
@kwwall , I'll write some unit tests for this, but I think this one can go in the next release.

sunnypav:
Sorry for the late reply when is the targeted release for the update ?

sunnypav:
Can i know if there is any release is planned for ESAPI. Also, if planned will this fix be considered for the same. Thanks

kwwall:
@sunnypav We have know _planned_ release, but that's one of the things that @xeno6696 and I still need to discuss. I'd certainly like to do a release before (say) July, but between now and 4/15, Uncle Sam beckons with his pesky taxes and unfortunately, that has to take priority.

@xeno6696, did you get a chance to write some additional unit tests for this? If so, I think we should add them at the same time that this gets committed. If there are some additional unit tests (including ones to make sure that this PR didn't break anything, like maybe testing something like '&aMp;', '&Lt;', and some accented cases and upper/lowercase phi, etc.) then in the meantime, I'd be more than happy to accept @sunnypav PR directly into the develop branch. That's not as good as an official production release, but at least it makes it easy for those building ESAPI from scratch.

Matt, I'll shoot you a separate email about ESAPI roadmap. I had something almost ready and then got distracted.

Thoughts?
-kevin

xeno6696:
My apologies @sunnypav, but it makes sense to me that the code writer provide tests for their own code.

kwwall:
@sunnypav Completely with @xeno6696 on this...unless it is for a PR request that is so trivial that no new test is needed, such as fixing a typo in an exception message, etc.

xeno6696:
My apologies for repeating myself 3 times @sunnypav .  I didn't mean to be aggressive... I'm still learning the review features of github.

xeno6696:
@kwwall Going to go ahead and pull this one in.  Already have failing unit test on the baseline for this one.

PR #380, opened 2016-06-21
Support of Cookie without maxAge set

mickilous:
The DefaultHttpUtilities and SecurityWrapperResponse add systematically a Max-Age to the HTTP Header when adding a cookie. Even if the cookie has a negative maxAge value (which is the default value).

So when adding a "session" cookie (without Max-Age), ESAPI add a Max-Age=-1 in the HTTP Header and the cookie will be discarded by the browser, because it is invalid (http://www.ietf.org/rfc/rfc2109.txt).
Wen the maxAge field of the cookie is negative, it should not be specified at HTTP Header level.

### DIFF
--- a/src/main/java/org/owasp/esapi/filters/SecurityWrapperResponse.java
+++ b/src/main/java/org/owasp/esapi/filters/SecurityWrapperResponse.java
@@ -124,7 +124,9 @@ public class SecurityWrapperResponse extends HttpServletResponseWrapper implemen
         // Set-Cookie:<name>=<value>[; <name>=<value>][; expires=<date>][;
         // domain=<domain_name>][; path=<some_path>][; secure][;HttpOnly
         String header = name + "=" + value;
-        header += "; Max-Age=" + maxAge;
+        if (maxAge >= 0) {
+            header += "; Max-Age=" + maxAge;
+        }
         if (domain != null) {
             header += "; Domain=" + domain;
         }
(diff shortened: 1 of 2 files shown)


framillien:
:+1: I want to use SecurityWrapperResponse, for a session cookie, in place of default servlet implementation, and fall to this issue too. Thanks.

xeno6696:
Hell with it. This one's in.

kwwall:
Probably no tests because that's hard to test without mock testing or
something like HttpUnit. ESAPI is pretty old and their widespread use was
probably not known at the time it started. Furthermore, the broad use of
singletons in ESAPI pretty much made mock tests worthless as a technique.

We should try to do better though if the effort isn't too great. If it is,
then we should consider dedicated manual code reviews instead.

Just my $.02,

-kevin
--

On Jul 16, 2017 12:11, "Matt Seil" <notifications@github.com> wrote:

*@xeno6696* approved this pull request.

As a rule, any changes to the baseline require unit test coverage. However
I just noticed that SecurityWrapperReponse doesn't appear to have a unit
test in the first place. @kwwall <https://github.com/kwwall> I'll create an
issue for this.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/ESAPI/esapi-java-legacy/pull/380#pullrequestreview-50214471>,
or mute the thread

PR #384, opened 2017-04-05
this commit fixes #385 changed visibility of method 'logSpecial(...)' from private to protected and #386 to avoid System.err output

### LINKED ISSUE #385: Method logSpecial in DefaultSecurityConfiguration is private and cannot be overriden by subclasses
To be able to change the behaviour of the method logSpecial() in DefaultSecurityConfiguration to e.g. put the output to an own Logger, the methods needs to be protected instead of private.

matthiaslarisch:
Changed the visibility of the method 'logSpecial(...)' to be able to override it to avoid the standard console output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ out/
 ciphertext-portable.ser
 ReferenceEncryptedProperties.test.txt
 test.out
-/pom.xml
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ out/
 ciphertext-portable.ser
 ReferenceEncryptedProperties.test.txt
 test.out
+/pom.xml
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
@@ -30,7 +30,8 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
             try {
                 return loader.getIntProp(propertyName);
             } catch (ConfigurationException e) {
-                System.err.println("Property not found in " + loader.name());
+                //System.err.println("Property not found in " + loader.name());
+                //FIXME: avoid using System.err (see issue #386)
             }
         }
         throw new ConfigurationException("Could not find property " + propertyName + " in configuration");
@@ -45,7 +46,8 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
             try {
                 return loader.getByteArrayProp(propertyName);
             } catch (ConfigurationException e) {
-                System.err.println("Property not found in " + loader.name());
+                //System.err.println("Property not found in " + loader.name());
+                //FIXME: avoid using System.err (see issue #386)
             }
         }
         throw new ConfigurationException("Could not find property " + propertyName + " in configuration");
@@ -60,7 +62,8 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
             try {
                 return loader.getBooleanProp(propertyName);
             } catch (ConfigurationException e) {
-                System.err.println("Property not found in " + loader.name());
+                //System.err.println("Property not found in " + loader.name());
+                //FIXME: avoid using System.err (see issue #386)
             }
         }
         throw new ConfigurationException("Could not find property " + propertyName + " in configuration");
@@ -75,7 +78,8 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
             try {
                 return loader.getStringProp(propertyName);
             } catch (ConfigurationException e) {
-                System.err.println("Property : " + propertyName + " not found in " + loader.name());
+                //System.err.println("Property : " + propertyName + " not found in " + loader.name());
+                //FIXME: avoid using System.err (see issue #386)
             }
         }
         throw new ConfigurationException("Could not find property " + propertyName + " in configuration");
@@ -86,12 +90,14 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
         try {
             loaders.add(createPropertyLoader(EsapiConfiguration.OPSTEAM_ESAPI_CFG));
         } catch (Exception e) {
-            System.err.println(e.getMessage());
+            //System.err.println(e.getMessage());
+            //FIXME: avoid using System.err (see issue #386)
         }
         try {
             loaders.add(createPropertyLoader(EsapiConfiguration.DEVTEAM_ESAPI_CFG));
         } catch (Exception e) {
-            System.err.println(e.getMessage());
+            //System.err.println(e.getMessage());
+            //FIXME: avoid using System.err (see issue #386)
         }
     }
 
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -252,7 +252,7 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     	this(DEFAULT_RESOURCE_FILE);
     }
     
-    protected final void setCipherXProperties() {
+    private void setCipherXProperties() {
 		// TODO: FUTURE: Replace by future CryptoControls class???
 		// See SecurityConfiguration.setCipherTransformation() for
 		// explanation of this.
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -252,7 +252,7 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     	this(DEFAULT_RESOURCE_FILE);
     }
     
-    protected void setCipherXProperties() {
+    protected final void setCipherXProperties() {
 		// TODO: FUTURE: Replace by future CryptoControls class???
 		// See SecurityConfiguration.setCipherTransformation() for
 		// explanation of this.
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -251,7 +251,8 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     public DefaultSecurityConfiguration(){
     	this(DEFAULT_RESOURCE_FILE);
     }
-    private void setCipherXProperties() {
+    
+    protected void setCipherXProperties() {
 		// TODO: FUTURE: Replace by future CryptoControls class???
 		// See SecurityConfiguration.setCipherTransformation() for
 		// explanation of this.
@@ -698,8 +699,8 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     	if (e != null) {
     		msg.append(" Exception was: ").append( e.toString() );
     	}
-	System.out.println( msg.toString() );
-	// if ( e != null) e.printStackTrace();		// TODO ??? Do we want this?
+		System.out.println( msg.toString() );
+		// if ( e != null) e.printStackTrace();		// TODO ??? Do we want this?
     }
 
     /**
@@ -710,7 +711,7 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
      * @param message The message to send to the console.
      */
     protected void logSpecial(String message) {
-	System.out.println(message);
+    	System.out.println(message);
     }
     
     /**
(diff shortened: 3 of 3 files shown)


kwwall:
First off, no pull requests accepted unless they are associated with a GitHub issue. Please create GitHub issues for these and reference the pull requests there. Otherwise it makes putting together release notes a lot more difficult. I'm also okay with changing the visibility of logSpecial() methods from private to protected so you can subclass them, but maybe a better question is WHY do you need to subclass this at all? Maybe there's a better way to accomplish what you need.

However, I am NOT okay with changing the visibility of setCipherXProperties(). That is called in the CTOR and you should NEVER allow methods called from a CTOR to be overridden. NEVER! That can result in trying to address partially constructed objects. It also can introduce security vulnerabilities through subclassing. Once I finish the changes to JavaEncryptor though, this kludge should no longer be necessary anyway.

So please plan accordingly or let me know what you are trying to accomplish and perhaps we can figure out a less risky way to accomplish it.

matthiaslarisch:
1. 

To avoid that the output is thrown to the console we subclassed the DefaultSecurityConfiguration and needed to override the to methods 'logSpecial()'. To be able to do that, this methods need to be protected instead of private.

2. 

To avoid, that we always get the messages that 'opsteam' and 'devteam' is not set in the System properties, which we don't need, we implemented an own constructor in our subclassed DefaultSecurityConfiguration to by pass the check for this two properties and for that, we need the setCipherXPorperties to be protected to be able to call the method from our constructor. Here is how the constructor look like:
`public MyOwnDefaultSecurityConfiguration() {
		super(new Properties());

		try {
        	loadConfiguration();
        	setCipherXProperties();
        } catch( IOException e ) {
	        logSpecial("Failed to load security configuration", e );
	        throw new ConfigurationException("Failed to load security configuration", e);
        }
	}`

3. 

Putting the pom.xml file to .gitignore was a mistake, I don't wanted to commit this file. I'm sorry for that.

kwwall:
[If this looks like crap when viewed from GitHub my apologies, but I
wanted to cross post this to the ESAPI-Dev mailing list.]

On Fri, Apr 7, 2017 at 2:37 AM, matthiaslarisch
<notifications@github.com> wrote:
>
> To avoid that the output is thrown to the console we subclassed the DefaultSecurityConfiguration and needed to override the to methods 'logSpecial()'. To be able to do that, this methods need to be protected instead of private.
>
> To avoid, that we always get the messages that 'opsteam' and 'devteam' is not set in the System properties, which we don't need, we implemented an own constructor in our subclassed DefaultSecurityConfiguration to by pass the check for this two properties

Are you referring to these two messages that come out on stderr?

    System property [org.owasp.esapi.opsteam] is not set
    System property [org.owasp.esapi.devteam] is not set

That's your entire reason for doing this? Okay. Wow. Seems a lot of
effort for so little return, but okay. Most people would just be happy
to ignore it, or better switch over to using ESAPI properties in this
new manner, but whatever.  Compared to all the (mostly useless) output
that most application servers spit out, this is a drop in the bucket.

Anyway, as previously explained, I have no issue with changing
logSpecial() from 'private' to 'protected'. While it is called from
the CTOR, did doesn't actuall aid in construction the object, so it
doesn't really matter. It's a no harm, no foul.

> and for that, we need the setCipherXPorperties to be protected to be able to call the method from our constructor. Here is how the constructor look like:
> `public MyOwnDefaultSecurityConfiguration() {
  super(new Properties());

  try {
      loadConfiguration();
      setCipherXProperties();
      } catch( IOException e ) {
          logSpecial("Failed to load security configuration", e );
          throw new ConfigurationException("Failed to load security
configuration", e);
      }
  }`
>

Well, except that I don't think that you do call setCipherXProperties().

As I previously mentioned, I am not going to allow a method that is
called from a CTOR to be overridden. There are two reasons for this.
One, doing so is simply bad practice: overriding any method called
from a constructor where that method sets instance variables could
potentially result in a partially constructed object. E.g., consider
if the derived method did not set those member instance variables.

Secondly, allowing a method called from CTOR in a base class to be
overridden in a class that has security implications (such as
DefaultSecurityConfiguration does) could allow attacker to bypass or
weaken security controls. (Some SAST tools flag even this sort of
thing as potential security concerns.)

So as I decribed previously, I will *NOT* accept any pull requests
that attempts to make setCipherXProperties() 'protected' or 'public'
as doing so could allow it to be used in a deliberately malicious
manner. Even if your proposed code is not doing this it doesn't mean
that is couldn't be abused in the future.

In hindsight, I think that the original ESAPI team should have
declared DefaultSecurityConfiguration as a 'final' class, but doing so
now would likely break someone's code and we try to avoid that as
well.

In the worst case, you could define your
'MyOwnDefaultSecurityConfiguration' class by simply copying
'DefaultSecurityConfiguration' and delete the use of
EsapiPropertyManager class which is where the these messages to stderr
are originating.

So, that's why I'm not going to allow setCipherXProperties() to be
overridden.  But I'm not seeing even why you think that you do. Here's
your code again:

    `public MyOwnDefaultSecurityConfiguration() {
        super(new Properties());

        try {
            loadConfiguration();
            setCipherXProperties();
        } catch( IOException e ) {
            logSpecial("Failed to load security configuration", e );
            throw new ConfigurationException("Failed to load security
configuration", e);
        }
    }`

I'm assuming that MyOwnDefaultSecurityConfiguration extends
DefaultSecurityConfiguration. If that is indeed the case, then

    super(new Properties());

will automatically call
DefaultSecurityConfiguration.setCipherXProperties() via the
DefaultSecurityConfiguration(Properties) CTOR. So, I think that all
you really need to do is, instead of passing in an empty Properties
object is to pass in (to the super() call) a Properties object that
has the "Encryptor.CipherTransformation" property value set to
whatever you want it to be. Then it should be picked up. I think you
should be able to do that in a static initializer. I've not tested
this and there may be more to all this than you are telling me. OTOH,
if you never intend to use ESAPI's encryption, it really doesn't
matter if this the instance variables that setCipherXProperties() sets
are set or not.

> Putting the pom.xml file to .gitignore was a mistake, I don't wanted to commit this file. I'm sorry for that.

No worries.

-kevin
-- 
Blog: http://off-the-wall-security.blogspot.com/    | Twitter: @KevinWWall

matthiaslarisch:
I understand your concern but what about setting the 'setCipherXProperties()' method to 'protected final'. Then this method could be called from a subclass but cannot be overriden and because this method has no parameters nothing could be manipulated by calling this method.

kwwall:
Well, if the ONLY reason you are wanting to do this is to silence these 2
messages to System.err, perhaps we really ought to deal with *that* issue
in a more direct fashion! So is it, or is there also some other reason you
want to subclass DefaultSecurityConfiguration???

-kevin
--

On Apr 10, 2017 02:09, "matthiaslarisch" <notifications@github.com> wrote:

I understand your concern but what about setting the
'setCipherXProperties()' method to 'protected final'. Then this method
could be called from a subclass but cannot be overriden and because this
method has no parameters nothing could be manipulated by calling this
method.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/ESAPI/esapi-java-legacy/pull/384#issuecomment-292858007>,
or mute the thread

matthiaslarisch:
That is REALLY the ONLY reason. We want to avoid any console outputs. That's why I subclassed the 'DefaultSecurityConfiguration' and overrode the 'logSpecial()' method and created an own Constructor to avoid the two 'System.err' outputs.

kwwall:
Hmm. Let me think about this a day or 2 and get feedback from others.

In general, I dislike hard coding output to either System.out or
System.err. I'm trying to assess whether we should do a broader, proper
cleanup of these (to allow them all to be redirected a specific file or
/dev/null, etc.) Or whether we should just address this the quick and dirty
way and just delete the offending line in EsapiPropertyManager?? Of course,
had Sun originally ​provided a way to dup file descriptors (e.g., see bug
ID JDK-4274770), we could have our cake and eat it too.

@xeno6696 - do you have any thoughts on this?

-kevin
--

xeno6696:
@kwwall 

I'd rather force the use of the logger or getting rid of the messages entirely.

kwwall:
Well, the logSpecial() method was there to fill a gap at the point where
ESAPI is still figuring out its properties, specifically, which logging
class is used is the one that is relevant here.

Things could /should be improved for those special cases IMO, but not sure
I want to address it now.

@matthiaslarisch Why don't you submit a PR with that 1 System.err.println()
commented out (along with a TODO or FIXME comment so I will remember to fix
it later & ref to the GitHub issue #) for EsapiPropertyManager and delete
your other PR and I can at least get it committed?

-kevin
--

jeremiahjstacey:
As another potential solution for the specific system, it is possible to override the default Sys out, in, and error streams for the specific application.

This tutorial shows how to configure the stream for the JVM.
http://tutorials.jenkov.com/java-io/system-in-out-error.html

Using this approach you still have the ability to pipe it to a file, or discard it entirely if you don't want it.  It leaves the information available for other users though, in the case where it provides value for debugging or analyzing an issue.

I'm not against the code change, and I fully support the update to remove the use of System streams in place of loggers.  I feel this solution could be a reasonable middle ground that can be applied on a case-by-case basis by the implementors.

matthiaslarisch:
I set the visibility of 'setCipherXProperties()' back to private.

xeno6696:
@kwwall I'm hesitant to merge this.  You clearly lead the charge on our behalf here, and I have a tendency to punt code changes that are relatively insignificant.  (No offense to the author, but I prefer hyper-verbose logging and choosing what I want to exclude when I want to exclude it.)  

If it were up to me, I'd punt this pull and focus on the removal of System.x calls.

kwwall:
@xeno6696 I'm inclined to agree. I'm going to email you off-GitHut to discuss it (because I think GitHub makes it too much of a PITA to do much more than top-post). We will make a decision and then I'll come back and follow up on this.

Rossi1337:
I suggest another approach to this. 
Currently the logSpecial writes via System.out.println

Maybe the easiest would be to introduce a static PrintStream variable "logStream" and initialize it with System.out

Then we provide a static setter to change this to an own print writer.

private static PrintStream  logStream = System.out;

public static setLogStream(PrintStream printStream) {
  logStream = printStream;
}

...

protected void logSpecial(String message) {
    	// System.out.println(message);
       logStream.println(message);
}

So we could set any other print stream and redirect the output to another logger or discard it.

Best regards,  
- Bernd Rosstauscher

kwwall:
@Rossi1337, I see the downside of this idea as two-fold:

1. What should ESAPI do if there is an `IOException` or an `NullPointerException` while attempting to write to the specified `PrintStream`? While the former conceivably _could_ happen with output to `System.out`, we at least know that this `PrintStream` has been properly initialized by the JVM so at least NPEs should be extremely rare.
2. Secondly it is not clear if this could potentially allow confusing interleaved output if an application switched between one `PrintStream` instance and another in their application. While unlikely to cause any harm (unless the `PrintStream` was directed as somewhere that ESAPI should never be writing, which is also a danger with this approach), I could see this becoming yet another topic for countless discussions on Stack Overflow, etc.

We are working on 2 approaches to this that we think will satisfy most developers. The first is that we have changed / added suitable `logSpecial(String msg)` and `logSpecial(String msg, Throwable t)` methods so that they are protected rather than private so that they an be overridden for these 4 ESAPI classes:
- `org.owasp.esapi.reference.DefaultSecurityConfiguration`
- `org.owasp.esapi.configuration.StandardEsapiPropertyLoader`
- `org.owasp.esapi.configuration,EsapiPropertyManager`
- `org.owasp.esapi.configuration.AbstractPrioritizedPropertyLoader`

This allows overriding these by sub-classing to do whatever you want. However, given that the original intent was simply to suppress `System.out` output (and that that seems the most likely reason this would be done as most DevOps seem to just simply redirect **ALL** `stdout` and `stderr` to the same place where Tomcat, JBoss, WebSphere, WebLogic Server, etc. output goes as part of the application server start up and just collect it with everything else), sub-classing seemed like a _lot_ of effort just to suppress this ESAPI output.

So, in addition to making all these `logSpecial()` methods **protected**, I am also changing all of these default `logSpecial()` methods to simply discard all output intended to `System.out` should the new `System` property "**org.owasp.esapi.logSpecial.discard**" be set to `true`. This allows one to simply suppress all such output from the default logSpecial() methods by doing something as simple as starting your JVM as:
`java -Dorg.owasp.esapi.logSpecial.discard=true ...`

We hope that between these two mechanisms, this will work for almost everyone.

kwwall:
In the end, after further analysis, I could not find a way that I was satisfied to make 'logSpecial' protected without introducing a potential security vulnerability. Therefore, I rewrote this PR and submitted it as PR #451 which I think will have the same effect.

I am therefore rejecting this PR but I hope that you will find theses 2 issues satisfactorily dealt with since your sole stated purpose was to suppress the output to System.err and System.out via logSpecial.

Issue #387, opened 2017-04-16
'mvn site' fails for FindBugs report, causing 'site' goal to fail


kwwall:
Maven produces the following error message on ESAPI's latest pom.xml (from the commit made on 4/16/2017):
`[INFO] All plugins with a version specified are using the latest versions.

[INFO]
[INFO] All plugins have a version specified.
[INFO]
[INFO] Project defines minimum Maven version as: 3.0
[INFO] Plugins require minimum Maven version of: 3.1
[INFO]
[ERROR] Project requires an incorrect minimum version of Maven.
[ERROR] Either change plugin versions to those compatible with 3.0
[ERROR] or update the pom.xml to contain
[ERROR]     &lt;prerequisites&gt;
[ERROR]       &nbsp;&nbsp;&nbsp;&nbsp;&lt;maven&gt;3.1&lt;/maven&gt;
[ERROR]     &lt;/prerequisites&gt;
[INFO]
[INFO] Require Maven 3.0.1 to use the following plugin updates:
[INFO]   org.codehaus.mojo:findbugs-maven-plugin ....................... 3.0.3
[INFO]`

I think there are at least 2 ways to address this. One is to require Maven version 3.1 or later. The other is to find a previous `org.codehaus.mojo:findbugs-maven-plugin` version that works with Maven 3.0. (The version that currently is specified is 2.5.5.)

I am somewhat reluctant to require Maven 3.1 as Maven 3.0 can even be a minor PITA to install (at least on Linux systems where the current deb or rpm package wants to pull in some ancient Maven 2.x version instead). E.g., on Linux Mint 17.2, an LTS version based on Ubuntu 14.04.2 LTS, I had to manually install Maven 3.0.5 to get it to work. I'm not sure everyone who wants to build ESAPI is up for that level of masochism, so I'd like to discuss the options.  We could keep the Maven 3.0 requirement instead of upping it to 3.1 if we can find out what version of FindBugs maven plugin works with Maven 3.0.

Let's discuss people's preferences!
-kevin

augustd:
3.1 was released in July 2013, so I don't think it is that onerous to expect developers to have a newer version than that! It was fairly simple to update my IDE to point to a newer version of maven (v3.3.9) and then update pom.xml to require 3.1: 

```
    <prerequisites>
        <maven>3.1</maven>
    </prerequisites>
```

I can send a pull request if you like...

augustd:
[...]

PR #388, opened 2017-05-03
Suppress CVE-2016-1000031 in dependency check

augustd:
See #387


kwwall:
Close per commit f012c40.

Issue #389, opened 2017-05-08
Provide an option for the encodeForLDAP method to not encode wildcard characters


JoelRabinovitch:
Hi,

Currently the encodeForLDAP method in the DefaultEncoder class encodes wildcard (*) characters.

While this is useful when trying to do a direct match on a value in an LDAP directory, you cannot use the method to find values in an LDAP directory using wildcards, while at the same time, escaping other potentially dangerous characters.  The scenario would be used when developing an LDAP query tool.

To resolve this issue, the encodeForLDAP method needs to be changed from:

// From:
    
    /**
     * {@inheritDoc}
     */
    public String encodeForLDAP(String input) {
        if( input == null ) {
            return null;    
        }
        // TODO: replace with LDAP codec
        StringBuilder sb = new StringBuilder();
        for (int i = 0; i < input.length(); i++) {
            char c = input.charAt(i);
            switch (c) {
            case '\\':
                sb.append("\\5c");
                break;
            case '*':
                sb.append("\\2a");
                break;
            case '(':
                sb.append("\\28");
                break;
            case ')':
                sb.append("\\29");
                break;
            case '\0':
                sb.append("\\00");
                break;
            default:
                sb.append(c);
            }
        }
        return sb.toString();
    }

to:

 // Start JavaDoc for Encoder Interface.
    
    /**
     * Encode data for use in LDAP queries. Wildcard (*) characters will be escaped.
     *
     * @param input the text to encode for LDAP
     *
     * @return input encoded for use in LDAP
     */
    
    // End JavaDoc for Encoder Interface.
    

    /**
     * {@inheritDoc}
     */
    public String encodeForLDAP(String input) {
        return encodeForLDAP(input, true);
    }

    // Start JavaDoc for Encoder Interface.
    
    /**
     * Encode data for use in LDAP queries. You have the option to escape wildcard (*) characters.
     *
     * @param input the text to encode for LDAP
     * @param escapeWildcards whether or not wildcard (*) characters will be escaped.
     *
     * @return input encoded for use in LDAP
     */
    
    // End JavaDoc for Encoder Interface.

    /**
     * {@inheritDoc}
     */
    public String encodeForLDAP(String input, boolean escapeWildcards) {
        if (input == null) {
            return null;
        }
        // TODO: replace with LDAP codec
        StringBuilder sb = new StringBuilder();
        for (int i = 0; i < input.length(); i++) {
            char c = input.charAt(i);

            if (c == '\\') {
[...]

PR #390, opened 2017-05-18
Issue #389

JoelRabinovitch:
Overloaded the encodeForLDAP method to provide the option of not
encoding wildcard (*) characters.

This would be used when doing queries against an LDAP directory using
wildcards, while at the same time, encoding other potentially dangerous
characters.


xeno6696:
@kwwall Please review this so we can close the encode for LDAP issue.  I haven't had time to look at the test case coverage.

kwwall:
Close issue #389

Issue #316, opened 2014-11-13
Deprecate current HttpUtilities.setRememberToken() and replace with one not requiring user password


meg23:
_From [kevin.w.wall@gmail.com](https://code.google.com/u/kevin.w.wall@gmail.com/) on November 26, 2013 13:27:30_

The current HttpUtilities.setRememberToken(String password, int maxAge, String domain, String path) interface takes a user's cleartext password. The reference implementation (DefaultHttpUtilities) creates an AES encrypted HTTP cookie that contains the username and password of the user to be remembered. Storing the password using reversible encryption is contrary to the stated corporate security policies of many companies and therefore using this as a technique to provide unattended login access should be discouraged if for no other reason than corporate policy will (or should) prevent this method from seeing widespread use.

Instead, a new replacement interface something like this:
HttpUtilities.setRememberToken(int maxAge, String domain, String path)
should be implemented to take advantage of CryptoToken which does not require a user's password to be made secure.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=311_

xeno6696:
@kwwall, is there a way we can "assign" this issue to me?  I'm willing to make this improvement.

kwwall:
@xeno6696 Matt - Go for it. Generally we have people either submit a pull request or attach a patch which one of us first reviews before we give someone commit access. That doesn't mean we doubt your mad skillz, but that's just been our practice because before we started that in Google Code we had (at one point) maybe a dozen people who had commit access but never submitted any code. Unfortunately, I don't think I can officially assign this issue to you until you get added to this project on GitHub as having commit access. (Well, let me put it this way, I tried and I was unable to assign it to your user id, so I'm speculating that is the reason why.)

xeno6696:
@kwwall do we want to just deprecate the old method and create a new one, or do we want to nukey nukey?

xeno6696:
The unit test for setRememberToken is actually partially commented out.  We never test it due to some dependency where the cookie is set via setSafeHeader.  

I don't know why that makes it untestable.

xeno6696:
@kwwall , I had to use a bit of reflection but I have a passing unit test for the replacement.  So, how did ESAPI typically test something like Cookies in an "integration-like" environment?  Our new cookie certainly gets set in the mock response, but before I push I wanted your input.

kwwall:
[...]

PR #393, opened 2017-06-18
Issue 316 -- updated code to account for httpOnly and Secure cookie …

xeno6696:
@kwwall -- I can't seem to find where the "deprecated" httpOnly exists from my original comment in #316

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -309,7 +309,7 @@
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>findbugs-maven-plugin</artifactId>
-                <version>2.5.5</version>
+                <version>3.0.4</version>
                 <configuration>
                     <findbugsXmlOutput>true</findbugsXmlOutput>
                     <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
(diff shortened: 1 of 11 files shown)


xeno6696:
@kwwall my apologies... github doesn't work like gerritt:  looks like this single pull request will also include the changes I just made for the search directory issue.  

I still need to write a series of unit tests to ensure that if those directory paths get changed again in the future, it'll cause the build to fail.

kwwall:
@xeno6696 No worries. Regarding the JUnit tests, it wouldn't surprise me if we have any tests at all trying to load ESAPI.properties from all the legal places so it's not surprising that we missed this the last time. But if you can think of an easy way to test this via JUnit, then go for it. I did look at the most recent code change though and it looks spot on.

-kevin

xeno6696:
@kwwall you're falling behind on reviewing all the commits here...

xeno6696:
@jeremiahjstacey I could use your code review expertise here.  I haven't committed professional code in awhile, want to make sure I'm not stupid.

xeno6696:
Per our conversation yesterday @kwwall I'm going to merge these changes in.  The last commits from yesterday created an Enum for the path search locations and added unit tests for when items in the enum get added/deleted and ensures that if those values ever change and the test isn't updated it'll break the build.

Issue #403, opened 2017-07-16
Refactor all "Magic Numbers" in the baseline to use ESAPI.properties configurations


xeno6696:
Also includes removing all deprecated security configuration calls.

xeno6696:
Using the regex `\(.*[0-9]+.*\)` to find the rest of these.

kwwall:
What about hex-encoded ones? I know there are a few in the crypto, but I
think most of those need to be constants, not parameters.

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/  |  Twitter:  @KevinWWall
NSA: All your crypto bit are belong to us.

xeno6696:
hardcoded numbers in crypto?  :-D

![image](https://user-images.githubusercontent.com/9502785/28747207-6b1a23ce-744e-11e7-9abd-06b7146dca4c.png)

kwwall:
LOL. Sssshhhh. You don't reveal magicians' secrets.

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/  |  Twitter:  @KevinWWall
NSA: All your crypto bit are belong to us.

kwwall:
Besides, obviously not *my* code. I would have used 0x4 instead. Hex is
obviously more magical. ;-)

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/  |  Twitter:  @KevinWWall
NSA: All your crypto bit are belong to us.

xeno6696:
Jokes aside, just to be clear, I maybe misnamed the issue.  I'm hunting for any numbers that ought to be configurable.. not necessarily wrapped into well-named constants.  _looks at the codecs_ yeah, I'm fine with char values in there...

xeno6696:
I've run into a host of nasty issues with unit tests after refactoring the crypto classes.  @kwwall , how much longer do you have on the crypto changes?

kwwall:
Undo,. Please undo. More later.

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/  |  Twitter:  @KevinWWall
NSA: All your crypto bit are belong to us.

xeno6696:
Well, in regards to your email, I was scratching my head that the crypto unit tests were failing because they were grabbing Strings from the recommended `SecurityConfiguration.getXProp()` instead of from the deprecated methods.  At any rate, duly rolled back.  So at this point the only changes that will be left are the magic numbers.  We can worry about getting rid of the deprecated methods later.  

[...]

Issue #292, opened 2014-11-13
jsessionid validator regex in esapi.properties not applicable to ids generated by tomcat


meg23:
_From [Alex.x86@gmail.com](https://code.google.com/u/Alex.x86@gmail.com/) on September 03, 2012 15:13:59_

From http://tomcat.apache.org/tomcat-6.0-doc/config/manager.html :

The length of session ids created by this Manager, measured in bytes, excluding subsequent conversion to a hexadecimal string and excluding any JVM route information used for load balancing. The default is 16.

The validator regex in Validator.HTTPJSESSIONID should be changed from

^[A-Z0-9]{10,30}$

to

^[A-Z0-9]{10,32}$

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=286_

PR #410, opened 2017-07-28
Issue #292 && Issue #403 -- Updated default regex size for jsessionid…

xeno6696:
… from 30 to 32, and refactored related classes to NOT use magic numbers.'

@kwwall -- I need to add a second commit.  It's a bridge too far I think to get the httpRequest object fully unit tested prior to the pending release, but we can write unit tests to make sure that we can pull back expected properties so that we don't run into runtime issues.  For all the new properties I've added in the past couple weeks, I'll make sure they're all tested in `DefaultSecurityConfigurationTest`.  I know it seems pedantic to test properties, we're still releasing our properties files to the public, and we need to make sure that if we fat-finger or change anything, that it won't break things for people.

### DIFF
--- a/src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java
+++ b/src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java
@@ -127,13 +127,13 @@ public class SecurityWrapperRequest extends HttpServletRequestWrapper implements
      */
     public String getContextPath() {
         String path = getHttpServletRequest().getContextPath();
-
+        SecurityConfiguration sc = ESAPI.securityConfiguration();
 		//Return empty String for the ROOT context
 		if (path == null || "".equals(path.trim())) return "";
 
         String clean = "";
         try {
-            clean = ESAPI.validator().getValidInput("HTTP context path: " + path, path, "HTTPContextPath", 150, false);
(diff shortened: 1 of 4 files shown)


kwwall:
@xeno6696 If you are trying to clean up ALL the magic #s, then I think it is reasonable to just test to make sure that we have default settings for our properties (although, note that we have a different ESAPI.properties for test vs. the production one so we aren't _really_ testing the version that we need).

If you are going to put these into `DefaultSecurityConfigurationTest`, I would put them all into the same test case and (ideally) make it data driven so that we can just read the property name and expected value from a separate file, or maybe 3 files...one for String, one for integer values, and one with byte arrays for values. But even if you don't do that, at least keep them altogether in a single test case, otherwise `DefaultSecurityConfigurationTest` will become extremely bloated. (Apologies for stating the obvious, but am in the middle of something else and too lazy at the moment to look at how you were doing this.)

xeno6696:
I did it the lazy way, but if it makes you feel better, I did exactly what you said.  I could just be too tired from a long day, but making this one data-driven is a little more than I'm willing to chew, although ESAPI.properties already gives us a base to work from.

kwwall:
You forgot to change the other ESAPI.properties. See comment under this commit for more info.

xeno6696:
My intention is to change the other file once we stabilize all changes for this release.

Issue #300, opened 2014-11-13
non-BMP characters incorrectly encoded


meg23:
_From [julian.r...@googlemail.com](https://code.google.com/u/107651864577259151853/) on March 01, 2013 11:36:39_

What steps will reproduce the problem? 1. run test case What is the expected output? What do you see instead? The test case encodes a non-BMP character, which internally is represented in java as two chracters, yet needs to be serialized as a single HTML character. What version of the product are you using? On what operating system? 2.0.1, win Please provide any additional information below. Test:

```
public static void main(String[] args) {
    String test = new String (new int[]{0x2f804}, 0, 1);
    System.out.println(test + " " + test.length());
    System.out.println(ESAPI.encoder().encodeForHTML(test));
}
```

Note: this problem has been mentioned over two years ago in http://ainthek.blogspot.de/2010/09/orgowaspesapicodecshtmlentitycodecjava.html but apparently hasn't been fixed.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=294_

xeno6696:
Even with the provided test case, this is very cryptic.  

Okay, so the character in question is [你](http://www.fileformat.info/info/unicode/char/4f60/index.htm)

What isn't so clear to me, is why the user would expect `&#x2f804;` instead of `&#x4f60;`

xeno6696:
Further in the second test case, 

```
        assertEquals( expected, htmlCodec.decode("&#194564;") );
```

From what I can tell, these entities also represent the same character.  

Digging up the blog post from this issue #, 

http://ainthek.blogspot.de/2010/09/orgowaspesapicodecshtmlentitycodecjava.html

Looks like @planetlevel had a small argument with this coder, and I'd like to get more context before deciding what to do with this issue.  At present, this seems to be a highly esoteric use case where folks should be using Unicode if they want desired results.

xeno6696:
Is #303 a duplicate of this?

xeno6696:
Aaaah... getting into some fun esoterism in code points and BMP.  

The core problem is in Codec.java:

```
	/**
	 * Encode a String so that it can be safely used in a specific context.
	 * 
	 * @param immune
	 * @param input
	 * 		the String to encode
	 * @return the encoded String
	 */
	public String encode(char[] immune, String input) {
		StringBuilder sb = new StringBuilder();
		for (int i = 0; i < input.length(); i++) {
			char c = input.charAt(i);
[...]

PR #413, opened 2017-08-06
Issue #300 -- Fixing ESAPI's inability to handle non-BMP codepoints.

xeno6696:


xeno6696:
@kwwall -- Do you think I should bring back the old `HTMLEntityCodec` as `LegacyHTMLEntityCodec` so users can have the ability to fall back to old behavior via configuration?

xeno6696:
I'm not sure how conservative we want to be with making changes. 

I'm confident the updated Codec does everything the old one did, and does more, but what a pain if we had to issue a quick point release.  (Though it would certainly get some folks noticing that development is finally active again!)

kwwall:
@xeno6696 Regarding your question about `LegacyHTMLEntityCodec`, it wouldn't be a bad idea..because if something were accidentally broken, at least there would be a fall-back and they could use that as a workaround. (That makes any emergency point releases a little bit less urgent if the code was accidentally broken.) But in addition to making that class avaiable, I recommend immediately marking the whole class as deprecated, noting that is replaced with a new version of `HTMLEntityCodec` that not encodes non-BMP and that it will disappear in a future release and thus should only be used in the event of a bug in `HTMLEntityCodec`.

Also, I will try to review your code at least by tomorrow night. I'm composing a TL;DR email to SourceClear at the moment. :)

kwwall:
Note: in my previous comment "this chunk" was meant to refer to the changes to Codec.java in commit 660054d.

kwwall:
Will provide more feedback / review tomorrow.

xeno6696:
> Am I correct in assuming that String.length() returns something that counts any non-BMP character as 2 characters? If not, I'm not sure I understand out the upper bound on the 'for' loop at line 67:
> offset < input.length()
> will work out correctly. Other than that though, this chuck looks good.

So I got the "Canonical" loop from this blog:

https://blogs.oracle.com/darcy/iterating-over-the-codepoints-of-a-string

Relevant snippet:  


Now, the
`char`
s need to be considered as possible members of a surrogate pair representing a single code point. Currently the canonical loop for this operation is:

```
String s = ...
for(int cp, i = 0; i < s.length(); i += Character.charCount(cp)) {
    cp = s.codePointAt(i);
    // Process cp...
}
```

Debugging purposes made me bury `offset += Character.charCount(point);`  as I wanted to prove to myself it works.  It appears that `String.length()` always returns the total number of visual characters, irregardless of how many chars it takes to represent them under the hood.  

From the `Character.charCount()` documentation:  

> Determines the number of char values needed to represent the specified character (Unicode code point). If the specified character is equal to or greater than 0x10000, then the method returns 2. Otherwise, the method returns 1. 
> 
> This method doesn't validate the specified character to be a valid Unicode code point. The caller must validate the character value using isValidCodePoint if necessary.

So... I need to add one more line of validation everywhere I'm about to append to a `StringBuilder`.

xeno6696:
> In HTMLEntityCodec, that complex test for illegal chars in/out of certain hex range...that used to return '&&#fffd;' but now returns the actual encoded char. Could that cause any backwards compatibility problems, such as 2 apps--one encoding and storing such a string in a DB or file--where used by different version of ESAPI to decode it. Seems unlikely someone would do that, but I have to ask about the edge cases.

That scares me a little because I got no unit-test failures in regards to this, and I would have expected it to light up like an x-mas tree.  But... searching the codebase I find no unit tests with an expected value of `&&#fffd;`

I will add more testing and roll in the other comment about a more complicated mixed BMP and Non BMP string in EncoderTest.

kwwall:
I ran through some of these really quick and also only looked at the diffs. Lots of stuff changed, so we need to make sure that we add comments to the ESAPI release notes that many of the signatures for the methods of classes in the `org.owasp.esapi.codec `package changed. I don't think that's a big deal. It wasn't really intended that people work with those directly (although I know some who do). If it becomes a problem (i.e., it breaks code and we get complaints) then can either use the legacy codec class or we can patch things up by adding some deprecated classes that will remove in 6 months to a year. (But for that reason, perhaps we should bump the minor #, despite my crypto / CVE related objections. (Less discuss it off GitHub.)  Also see my email about the author annotation. You at least should probably add yourself.

Anyhow, I'm good if you want to do the honors and merge this. I can never figure out which option is the best. (GitHub is suggesting "squash and merge", but also offers "create a merge commit" and "rebase and merge".) You know the pros and cons of each better than me so I will let you do the merge, but I approve of you changes.

xeno6696:
Scary.... no... not quite done yet.  Just noticed that the main body of the  `HTMLEntityCodec` was using the `Character` signature and the correct `int` version was mostly commented out.  I have a few more unit tests to clean up.

xeno6696:
Ah yes @kwwall -- You caught that with your comment about `ufffd` earlier.

xeno6696:
@kwwall -- Your review approval sort of missed code here.  Take a sharp eye.  

If you agree with this I'll merge this series to the baseline.

xeno6696:
>  Guess that's the right behavior, but it would be so much more hilarious if you could return the string for U+1F4A9, which is the Unicode character for the poop emoji. (Hey, where'd that pile of steaming shit come from? :)

```
 ROFL:ROFL:ROFL:ROFL
         _^___
 L    __/   [] \
LOL===__        \
 L      \________]
         I   I
        --------/
```

xeno6696:
@kwwall  

As a wrap-up task... do you think that we should possibly warn that if data in a string is not a valid code point that we deliberately destroy that data?  Or do you think this can be implied simply because if it's not a code point... you shouldn't be putting it into a `String?`

kwwall:
No, it definitely would be good to indicate in the Javadoc that non-valid point codes are ignored. (Especially needed if these can be put into a String in the first place, which I assume must be possible unless String CTORs do the same thing and silently skip over them.) At first I thought maybe we should throw an unchecked exception or something, but then realized that could be used as a DoS attack of sorts since exception handling is much more expensive in CPU usage than regular Java code and usually we are in a case where we are handling tainted user from and end user so no point in making attacker's lives any easier.

xeno6696:
I updated it a bit further--non UTF code points were silently destroyed in the old behavior by all codepoints being cast to `char` before being appended.  `Character` based codecs will still follow this behavior.

PR #421, opened 2017-08-27
Adding mocking frameworks for testing.  Also added joda-time in preparation for date fixes.

xeno6696:


kwwall:
@xeno6696 Long overdue; thanks. But all those singletons make mock testing much less useful than it could otherwise be.

xeno6696:
Not for the singletons as much as for getting the filters and other j2ee implementation under test

PR #426, opened 2017-11-06
Fix configuration loading: Use value instead of constants also update vulnerable dependencies

NiklasMehner:
The existing code uses enum names instead of the values to try to load ESAPI.properties.


kwwall:
@NiklasMehner In the future, my preference would be that you create separate GitHub issues for separate problems. That makes it much easier for us to create a coherent set of release notes as well as making it easier to accept one of the commits independent of the others, without having to alter the description of the GitHub issue.

Issue #135, opened 2014-11-13
SecurityWrapperRequest#getQueryString() decodes percent escapes


meg23:
_From [schal...@darkmist.net](https://code.google.com/u/106366406945487233817/) on May 15, 2010 21:42:05_

What steps will reproduce the problem? 1. Wrap a request that contains a &#37; escape in the query string
2. Call getQueryString()
3. note that percent escapes have been unescaped What is the expected output? What do you see instead? This is debatable. The problem here is that unescaping &#37; changes the
meaning of the query string. In normal form encoding
name1=value1&name2=value2, equals and ampersand must be escaped in names
and values. However, the unescaping that getQueryString is performing would
cause the values not to parse correctly. Please use labels and text to provide additional information. It needs to be decided what is best to do here. The purpose of this routine
makes canonicalization and validation difficult at best because it bypasses
the normal decoding into parameters that is done internally by the request.
I have seen urls with other meta characters in the query string than those
done for form encoding though I don't know what the specification says
about such.

My current thought on the issue is that this routine is generally expected
to return a raw value. As such it would probably be best just to validate
that the characters contained therein are valid characters for a query
string and not go farther. I believe this would just be URL chars without
the '#' but I would need to test.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=125_

xeno6696:
@jeremiahjstacey this is where you should be committing your work against.

jeremiahjstacey:
What do you want this to do?  It's catching the ValidationException and swallowing it.  I can either wrap it in a runtime exception and rethrow, or log it.  If it's logged, it will not change the blank string returned to the client. If the logging route is chosen, I would appreciate confirmation that the blank string is desired.  Otherwise, I think the only other option I have is to return the original Input reference.

I am of the opinion that this is an exception case, and a Runtime exception should be thrown, allowing the context to escape to the execution container.   I intend to move in this direction, but it will be easy enough to re-route if needed.

There are other places in the SecurityWrapperRequest where similar ValidationExceptions are being caught.  In some cases they appear to be logged, in others they are swallowed.  Should the decision made for resolving this instance be applied to those other locations as well?

kwwall:
[...]

PR #431, opened 2018-02-19
Jstacey 135

jeremiahjstacey:
This is related to the github issue #135 

In the course of discussion for this issue, it was determined that the issue as filed is intended functionality.  However, through that course we found that the behavior and intent of the tests were not entirely clear.  These changes should assist in the clarity of the SafeRequestTest by isolating the tests for the SecurityWrapperRequest implementation from the codec behavior.

Also provided is  a test-scope refactor of the implied coverage for the PercentCodec provided by the previous SafeRequestTest implementation.  Parameterized tests have been provided for asserting the PercentCodec support for both String and Character variations.  A negative test case was added for UTF16 character encoding, which is pending full implementation on that codec.

I believe this work should resolve issue 135.


kwwall:
Approve your commits as long as copyright and license statement added to all new files, although other changes are suggested. See comments I just emailed to you. (I can attach them here for completeness if you think that is a good idea.)

xeno6696:
@kwwall Assuming you're fine with the most recent push I'm ready to merge.

kwwall:
Isn't it {@code ...} rather than @{code ...} ??? Maybe that's what you
meant though.

-kevin
--

jeremiahjstacey:
After doing some  (limited) research on the use of {@code} it would not let me represent the javadoc the way that I wanted.  Other formatting suggestions were using <pre> tags and other nesting for formatting.  The change to <code> felt like the most succinct solution that rendered the way I had originally intended when looking at the documentation output (in my ide).

PR #436, opened 2018-06-19
Add unicode letter/number support for fileName validation

tom-leahy:
Updating file and directory validation regex to allow unicode letters/numbers

### DIFF
--- a/configuration/esapi/ESAPI.properties
+++ b/configuration/esapi/ESAPI.properties
@@ -1,6 +1,6 @@
 #
 # OWASP Enterprise Security API (ESAPI) Properties file -- PRODUCTION Version
-# 
+#
 # This file is part of the Open Web Application Security Project (OWASP)
 # Enterprise Security API (ESAPI) project. For details, please see
 # http://www.owasp.org/index.php/ESAPI.
@@ -102,12 +102,12 @@ Authenticator.AbsoluteTimeoutDuration=120
 # following code to canonicalize data.
 #
 #      ESAPI.Encoder().canonicalize( "%22hello world&#x22;" );
-#  
+#
 # Multiple encoding is when a single encoding format is applied multiple times. Allowing
 # multiple encoding is strongly discouraged.
 Encoder.AllowMultipleEncoding=false
 
-# Mixed encoding is when multiple different encoding formats are applied, or when 
+# Mixed encoding is when multiple different encoding formats are applied, or when
 # multiple formats are nested. Allowing multiple encoding is strongly discouraged.
 Encoder.AllowMixedEncoding=false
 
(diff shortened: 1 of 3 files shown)


kwwall:
@tom-leahy : We are currently considering if we want to make this the default, but my gut reaction is that I would prefer to follow the principle of least surprise or just leaving the defaults for these as-is and adding these new regexes supporting Unicode letters and numbers as a commented out option where developers can then easily choose only if they wish. While I realize we could note it in the release notes, many developers do not carefully read the release notes and them may be surprised when their application (say) goes from only accepting file names with ASCII characters to accepting all Unicode letters and numbers. I hope to have a decision within the week regarding this.

tom-leahy:
@kwwall - thanks for the review/update and that makes sense to me, please let me know. Another approach may be for me to add to the method overloading on isValidFileName/isValidDirectoryPath to indicate which pattern you wish to use (i.e. boolean allowUnicode) and just grab the existing or the new unicode inclusive pattern but make the default to use the old pattern.

kwwall:
@tom-leahy - Sorry this has taken so long, but one of my colleagues was on an extended road trip and just returned this morning.

I reached out to the two developers who are working with my on ESAPI as well as to two former project leaders. Their reactions ranged from "I wouldn't recommend changing the details" to "we should definitely not do this". However, none of them objected to leave it in the ESAPI.properties file as a commented out property to show folks how to do this.

Now, as you have seen, we have 2 ESAPI.properties, One for specifically for testing that has the encryption key set to a default, and a few other very minor changes. I prefer to leave them as close as identical as possible.  However, if I leave the properties for Validator.FileName and Validator.DirectoryName "as-is" in the test version of ESAPI.properties, your new test cases that check for Unicode letters in file names will fail. Now there is a way around this. I wanted to ask you if you would be willing to do it. We have SecurityConfigurationWrapper class under src/test/java/org/owasp/esapi that we use for special cases such as this...to temporarily use different properties. We are using it in ExecutorTest.java and we really should be using it in ValidatorTest.java as well. Would you be willing to rewrite your JUnit test for ValidatorTest.java to use SecurityConfigurationWrapper? If so, that would allow us to keep the official and test versions of ESAPI.properties in sync as much as possible. (You may have already noticed the FIXME comment in that file.) If you can do that, that would be great. Otherwise, I would just suggest suggest that you comment out the code and submit a new PR with everything appropriately commented out. Hope this is understandable. I'm in a hurry because I have to take my son to work.

kwwall:
@tom-leahy Sorry. Forgot to mention one thing. I am against adding new methods. IMO, most of our interfaces are already too bloated and this one for the Validator certainly is. Now if Java supported default arguments in the way that C++ did, we could add something like this without the bloat, but unfortunately it doesn't. Anyhow, wanted to address that.

tom-leahy:
@kwwall - sorry for the delay in response, I just got back from vacation. I took a crack at using the SecurityConfigurationWrapper to test the regex patterns, please take a look when you get a chance.

kwwall:
Close #436

Issue #129, opened 2014-11-13
Add Logging support for SLF4J


meg23:
_From [ps.gr...@gmail.com](https://code.google.com/u/114075484593062071800/) on April 27, 2010 03:09:02_

SLF4J is the next generation logging framework replacing commons logging and 
simplifies the complicated class loading used by commons logging(which normal 
breaks in Websphere). You can use a bridge to intercept log4j calls and 
direct these to SLF4J but feel it would be better to natively support SLF4J 
out of the box in ESAPI.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=119_

zhangshifeng:
@CarvalhoLeonardo and @jonesde has done [some work](https://github.com/moqui/moqui-framework/tree/master/framework/patches/esapi) on this issue, please give some advice.

jeremiahjstacey:
Just to clarify the intent of this issue, are we looking to replace/upgrade the existing log4j to slf4j or simply add slf4j along side the log4j capabilities?

xeno6696:
It should be replace.  Though, you get log4j compatability for free with slf4j.

kwwall:
As long as we keep the ESAPI logging interfaces the same, I think we can
just support SLF4J and J.U.L. and get the log4j support via SLF4J.

@xeno6696, @chrisisbeef, or @planetlevel -- Do you think that is an
accurate assessment?

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/  |  Twitter:  @KevinWWall
NSA: All your crypto bit are belong to us.

xeno6696:
@jeremiahjstacey did some sleuthing.  And for a moment I was in agreement that because most implementations seem to offer automatic CRLF injection protection...

Except java.util.logging, and log4j.  

The reasoning behind changing ESAPI to use slf4J, is that we automatically gain access to defending *all* logging frameworks in current use.  JUL, log4j, and logback all have common slf4j bridge libraries[, with 4/5 github projects using slf4J as their primary interface.  ](http://blog.takipi.com/is-standard-java-logging-dead-log4j-vs-log4j2-vs-logback-vs-java-util-logging/)

[...]

PR #446, opened 2018-09-25
Issue #129 SLF4J Logging Structures

jeremiahjstacey:
I believe this work is completed and functional.


xeno6696:
@kwwall There's only one conflict here, looks like it's with a change you made.  Not sure which one is the right one to pick?

kwwall:
I'll take a look.

-kevin
--

kwwall:
@xeno6696 We should keep the changes that I made in **EncryptorTest.java**. That agrees how the JUnit ought to operate based on observed actual behavior. So keep my changes and discard @jeremiahjstacey  changes for **EncryptorTest.java**.

jeremiahjstacey:
I agree. I think I had committed this during our discussion of the EncryptorTest behavior before I made the branch for that discussion.  I should have undone this change to this branch before I made the pull request.

PR #456, opened 2018-11-07
fix bug not found in classpath when run with jetty or tomcat

JasperXgwang:
jetty or tomcat use self classloader, ClassLoader.getSystemResource use jvm classloader


kwwall:
@JasperXgwang If we are to consider this PR, then we will insist on that you create a GitHub issue and associate this PR with it. In the GitHub issue, try to clearly describe what problem your PR is addressing and just as important how this will work in other environments besides just Tomcat and Jetty. That is the only way that we will accept this PR because we (humans following ESAPI) track things mostly through GitHub issues that than just commits and git commit logs. Thanks.

JasperXgwang:
@kwwall  ok, Thanks

xeno6696:
In the future @JasperXgwang, make sure to include the issue number in commit messages in future PRs... that way github automatically closes issues.

xeno6696:
Fixes issue #456

Issue #30, opened 2014-11-13
ESAPIFilter in RI should allow login page destination to be configured


meg23:
_From [chrisisbeef](https://code.google.com/u/chrisisbeef/) on July 24, 2009 03:50:59_

Rather than force uses of the RI to have their login page be located at
/WEB-INF/login.jsp we should allow configuration of where the users should
be forwarded in the event that a forward or redirect needs to occur to log
the user in. 

We should be able to add a line to the ESAPI.properties file which allows
the user to specify where login requests should be sent.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=20_

kwwall:
Note that PR #459 which closed this was identical to PR #460 which was rejected (as duplicate).

PR #459, opened 2018-12-20
Eclipse setup updates

simon0117:
I have loaded the project into Eclipse 2018-12 and resolved a maven lifecycle error generated by m2eclipse (comment in pom.xml refers to predecessor q4e) by just updating the plugin version. Also added Eclipse default bin dir to .gitignore.

PR #460, opened 2018-12-20
ESAPIFilter in RI should allow login page destination to be configured #30

simon0117:
I know I've messed up the pull request because of a previous pull request, just let me know what to do.

Regarding issue #30 I added "loginPage" filter init parameter in ESAPIFilter. I noticed though that there are other hardcoded pages such as WEB-INF/index.jsp on line 105, so you may want to create another issue for that (if there isn't one already, I haven't checked)

It's not perfect code because of the literal string "loginPage" in the middle of the code, but it should work fine.

I could not find any documentation to update to document this new "hidden feature" of ESAPIFilter. Unless this is a reference implementation (is that what RI means?) and not meant to be used in applications? Then why is it compiled in the library? I was unable to find (or create) appropriate unit tests either. In other words I wrote the code, mvn test keeps on giving me that known error, and I am unable to test if my code breaks anything or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.iws
 *.eml
 out/
+bin/
 
 # Leftover test files
 ciphertext-portable.ser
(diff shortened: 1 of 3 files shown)


kwwall:
There is no need for this PR. Apart from your different comments, it is identical with PR #459 which I already merged.

Issue #360, opened 2016-02-01
AuthenticatorTest.setCurrentUser() periodically fails.


kwwall:
There is an apparent race condition either in this test itself or in the FileBasedAuthenticator class that is used (indirectly) in the test. It seems to be failing about 20-30% of the time on my laptop (a dual core i5); YMMV.

@jeremiahjstacey and @xeno6696 are have been looking into this. Since no one is likely using FileBasedAuthenticator in any enterprise production scenario (if they are, they really need to re-evaluate that decision!), I am going to go ahead and release ESAPI 2.1.0.1 even if we do not have a code fix for this issue by the time of the release. If not, I will mention it in the release notes.

xeno6696:
Wow @jeremiahjstacey it was TWO years ago!!!!

kwwall:
I just recently got a failed test case that seems related to this:


testGetCurrentUser(org.owasp.esapi.reference.AuthenticatorTest)  Time elapsed: 0.156 sec  <<< ERROR!
org.owasp.esapi.errors.AuthenticationAccountsException: Account creation failed
        at org.owasp.esapi.reference.FileBasedAuthenticator.createUser(FileBasedAuthenticator.java:240)
        at org.owasp.esapi.reference.AuthenticatorTest$1.run(AuthenticatorTest.java:221)
        at java.lang.Thread.run(Thread.java:745)

kwwall:
Changing this to a low priority because no one in the right minds should be using the toy implementation of `FileBasedAuthenticator`.

kwwall:
**Note**: I set the label 'help wanted' on this issue not because this is a good first issue, but rather because we could use some assistance on it as we cannot reproduce this on demand and thus are unable to debug it.

kwwall:
Closed via PR #467

PR #467, opened 2018-12-30
AuthenticatorTest Stability #360

jeremiahjstacey:
Several changes that appear to improve the stability of the overall test.

First, broke testSetCurrentUserWithRequest into five uniquely named tests.
One of these is used to verify a user with an expired account cannot log
in.  Fix here is to explicitly set the expiration time in the past to
prevent the implementation from validating in the same millisecond.

Second, updated the internal runnable in testSetCurrentUser to generate a
more complex password.  Found that the previous password would
occasionally fail the FileBasedAuthenticator password strength check. I
assumed it was because the strength value was being calculated as either 8
or 16.  Stregth would evaluate to 8 if the 8-character string was composed
of only one unique type of character (upper/lower/number/special), 16 if
there were only 2 unique types. By setting the string length to 17 we
explicitly avoid this potential failure in testing.

Third, I had implemented a section of code in the tear-down behavior that
cleared the userDB file and used Whitebox to null the 'userDB' attribute
in the FileBasedAuthenticator instance.  Even with that I was still seeing
issues in testSetCurrentUser where the test would fail because the user
account would already exist.  To account for this, I modified all user
names to be randomly generated Strings which will avoid the test
unintentionally creating more than user by the same name.

Finally, in testGetCurrentUser there was a value being tracked for the
user compare 'result'.  Though part of the tests' validation, it would not
have failed the test since it was being checked within the Runnable
context and not being propagated back to the test execution scope.  I
modified the implementation to add the boolean 'result' check to the
ErrorCollector, which can fail the test if a mismatched user reference
occurs.

With these changes I can run this test in isolation 100/100 times
successfully.

I was able to run mvn clean test 500 times against the full baseline with 100% success.  (I used a shell script and grepped the results).  I believe these changes will decrease the intermittent failures from AuthenticatorTest.

### DIFF
--- a/src/test/java/org/owasp/esapi/reference/AuthenticatorTest.java
+++ b/src/test/java/org/owasp/esapi/reference/AuthenticatorTest.java
@@ -24,7 +24,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.junit.Assume.assumeTrue;
 
-import java.util.Date;
+import java.util.Calendar;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Semaphore;
@@ -33,6 +33,7 @@ import java.util.concurrent.TimeUnit;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.hamcrest.core.IsEqual;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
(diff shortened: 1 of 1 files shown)


kwwall:
I pulled down your commit for AuthenticatorTest.java and applied it to the
latest version of the 'develop' branch on my Linux Mint system and am still
getting the same failure. I'm getting slightly less than before (maybe 35%
compared to 40% before). Will investigate further tomorrow to make sure
that I didn't screw something up, which I'm hoping I did.

-kevin
--

PR #468, opened 2019-01-11
Date validation rule 299

jeremiahjstacey:
These commits offer a possible solution to issue #299.
Summary of changes:
  * DateValidationRule API extended to include a new sanitize method which accepts a ValidationErrorList
  * DefaultValidator updated to use the DateValidationRule sanatize method
  * Tests around implementation including sample strings from 299 and 258

The changes to DateValidationRule make a distinction between the getValid and sanitize API's.  getValid will make the best possible effort, given configuration and formatter, to determine if any date can be parsed from the input string with no guarantee of safety or accuracy of the input.  This route relies soley on the backing DateFormat for success.

sanitize extends the getValid behavior by attempting validate that the input can be considered 'safe' for application use.

### DIFF
--- a/src/test/java/org/owasp/esapi/reference/validation/BaseValidationRuleTest.java
+++ b/src/test/java/org/owasp/esapi/reference/validation/BaseValidationRuleTest.java
@@ -15,87 +15,274 @@
  */
 package org.owasp.esapi.reference.validation;
 
-import junit.framework.Test;
-import junit.framework.TestCase;
-import junit.framework.TestSuite;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.CALLS_REAL_METHODS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
 
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.hamcrest.Matcher;
+import org.hamcrest.core.Is;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.owasp.esapi.Encoder;
+import org.owasp.esapi.ValidationErrorList;
 import org.owasp.esapi.errors.ValidationException;
 
-public class BaseValidationRuleTest extends TestCase {
+public class BaseValidationRuleTest {
+    /**Static Test Data.*/
+    private static final String STR_VAL="";
+    private static final String EX_MSG="Expected Failure Message from " + BaseValidationRuleTest.class.getSimpleName();
+    @Rule
+    public ExpectedException exEx = ExpectedException.none();
 
-	/**
-	 * Instantiates a new base validation rule test.
-	 * 
-	 * @param testName
-	 *            the test name
-	 */
-	public BaseValidationRuleTest(String testName) {
-		super(testName);
-	}
+    private ValidationException testValidationEx = new ValidationException(EX_MSG, EX_MSG);
 
-	/**
-	 * {@inheritDoc}
-	 * 
-	 * @throws Exception
-	 */
-	protected void setUp() throws Exception {
-		// none
-	}
+    private BaseValidationRule uit = mock(BaseValidationRule.class, CALLS_REAL_METHODS);
+    @Test
+    public void testCtrNullTypeName() throws Exception { 
+        String typename = null;
+        BaseValidationRule rule = mock(BaseValidationRule.class, withSettings()
+                .useConstructor(typename)
+                .defaultAnswer(CALLS_REAL_METHODS)
+                );
+        assertNull(rule.getTypeName());
+    }
 
-	/**
-	 * {@inheritDoc}
-	 * 
-	 * @throws Exception
-	 */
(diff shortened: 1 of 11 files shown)


src/test/java/org/owasp/esapi/reference/validation/BaseValidationRuleTest.java
```diff
@@ -15,87 +15,274 @@
...
+        exEx.expect(NullPointerException.class);
+        String myString = "AAAGaaadBBB12345*";
+        Set<Character> whitelist = null;
+        uit.whitelist(myString, whitelist);
+    }
+
+    @Test
+    public void testWhitelistSetExtendedCharacterSets() {
+        String myString = "𡘾𦴩<𥻂";
+        //This was an issue wiht HTMLEntityCodec at one point, so it seems like we should be doing this test.
```
jeremiahjstacey: @xeno6696 Since you worked on the HTMLEntityCodec I thought you may have context into this test.  
This test is currently marked as Ignored, but it feels like something that should be verified.  
Interested in your thoughts.

xeno6696: I see you got the string from the HTMLEntityCodecTest...  I'm not sure why you didn't copy over the actual test....

xeno6696:
```
	@Test
	public void testEntityDecoding(){
		assertEquals("<", codec.decode("&lt;"));
        assertEquals( "<", codec.decode("&LT"));
        assertEquals( "<", codec.decode("&lt;"));
        assertEquals( "<", codec.decode("&LT;"));
	}
	
	@Test
	public void test32BitCJK(){
		String s = "𡘾𦴩𥻂";
		String expected = "&#x2163e;&#x26d29;&#x25ec2;";
		String bad = "&#xd845;&#xde3e;&#xd85b;&#xdd29;&#xd857;&#xdec2;";
		assertEquals(false, expected.equals(bad));
		assertEquals(expected, codec.encode(new char[0], s));
	}
	
	@Test
	public void test32BitCJKMixedWithBmp(){
		String s = "𡘾𦴩<𥻂";
		String expected = "&#x2163e;&#x26d29;&lt;&#x25ec2;";
		String bad = "&#xd845;&#xde3e;&#xd85b;&#xdd29;&#xd857;&#xdec2;";
		assertEquals(false, expected.equals(bad));
		assertEquals(expected, codec.encode(new char[0], s));
	}
	
	@Test
	public void testDecodeforChars(){
		String s = "!@$%()=+{}[]";
		String expected = "!@$%()=+{}[]";
		assertEquals(expected, codec.decode(s));
	}
	
	@Test
	public void testMixedBmpAndNonBmp(){
		String nonBMP = new String(new int[]{0x2f804}, 0, 1);
		String bmp = "<a";
		String expected = "&lt;a&#x2f804;";
		String input = bmp + nonBMP;
		assertEquals(expected, codec.encode(new char[0], input));
	}
```

xeno6696: obviously that very first test has a couple of mistakes that we missed, but these are the core tests.

jeremiahjstacey: In this case we're not encoding/decoding.  We're stripping characters from a string. 
I believe I have reasonable tests for the UTF-8 characters, but the utf-16 gave me some heartburn.

I think I tried to remove one of the characters from the origin string and ended up creating a test failure.  Since this was ancillary to the fix I was chasing I used the fail clause to point out that I wasn't sure what exactly I was doing, or trying to do,  so I could follow up on it later.

I think I was writing a bad test, or I had bad expectations with the character whitelist behavior.  I can take another crack at it.  Looking at it again, I think I need to whitelist the character pair that make up the symbol display.  That may suffice for showing that utf-16 can be whitelisted through BasicValidationRule.

I don't think the other tests add value here, which is why I only pulled this data set.

jeremiahjstacey: Yeah, I probably was not treating the utf-16 correctly.  This implementation shows my desired intention:

 ```
@Test
    
    public void testWhitelistSetExtendedCharacterSets() {
        String myString = "𡘾𦴩<𥻂";
        //(55365 56894) (55387 56617) 60 (55383 57026)
        Set<Character> whitelist = new HashSet<>();
        whitelist.add((char) 60);
        whitelist.add((char) 55387);
        whitelist.add((char) 56617);
        String result = uit.whitelist(myString, whitelist);
        assertEquals("𦴩<", result);
    }
```

jeremiahjstacey: This test is added to the pull request.

kwwall:
@jeremiahjstacey Wasn't this issue #299 the one that we thought that @xeno6696 added a dependency for joda-time:joda-time:2.10 in the pom.xml but then you found out that we really didn't need it? IIIRC, I don't think we are using it anywhere so could you add a commit to this PR that would delete that dependency from the pom.xml? (Or if we are using it for JUnit tests now, just change its scope to 'test'.) Thanks!

jeremiahjstacey:
@kwwall , JodaTime Dependency was cleaned up in one of my first commits to ensure I wasn't building on an incorrect assumption:
> Dependency Cleanup: JodaTime
> 
> Removing JodaTime dependency from Pom.  It had been introduced to
> potentially address date/time parsing issues, but had not yet been applied
> to the baseline.
> https://github.com/ESAPI/esapi-java-legacy/pull/468/commits/d9aa1a8c373d1755839b8772959d08740d8ef1dc

Issue #37, opened 2014-11-13
RandomAccessReferenceMap.update() can randomly corrupt the map


meg23:
_From [cyounk...@gmail.com](https://code.google.com/u/110242786105136439221/) on August 06, 2009 15:33:16_

What steps will reproduce the problem? You have a ARM set up like this:
   direct1 <=> "AAAA"

You do this:
newSet = new Set()
newSet.add(newDirect)
newSet.add(direct1)
ARM.update(newSet)

update() clears the old mappings, but tries to preserve them if the same
direct objects are referenced in the new set. Here's how this could go awry:
- update() clears both maps
- update() can't find newDirect in the old mapping, so it calls
  getUniqueRandomReference(), which randomly returns "AAAA". It checks that
  "AAAA" is not in the current mapping, and it's not (It was just cleared).
- update() sets the mapping newDirect <=> "AAAA"
- update() then looks at direct1, sees that it is in the old mapping, and
  sets it in the new mapping direct1 <=> "AAAA", corrupting the indirect to
  direct mapping which is now "AAAA" -> direct1

How to fix it:
- Change getUniqueRandomReference to take in a list of invalid indirect 
  keys and ensure the generated one is not in that list
- Preserve the old indirect to direct map in update()
- Pass in the keys of the old indirect to direct map to
  getUniqueRandomReference What version of the product are you using? On what operating system? SVN revision 574 .

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=27_

kwwall:
Closed via PR #469

PR #469, opened 2019-01-13
AbstractAccessReferenceMap Issues #37 #329

jeremiahjstacey:
Potential resolutions for two issues in AbstractAccessReferenceMap.

First, handling #329 I've provided a multi-threaded unit test which replicates the condition consistently.  The offered solution is to add *synchronized* keyword to all public API of the AbstractAccessReferenceMap.

Second, #37 Provided a test which shows that in cases where subclasses return the same indirect key multiple times it can lead to corruption during update.  Logic updates were made to the implementation to address.


kwwall:
Close issue #37 
Close issue #329

Issue #286, opened 2014-11-13
JavaLogFactory not thread safe


meg23:
_From [mike.til...@yahoo.com](https://code.google.com/u/109171368508383610253/) on July 26, 2012 10:07:44_

What steps will reproduce the problem? for (int i = 0; i < 5000; ++i) {
            new Thread(new Runnable() {
                    public void run() {
                        ESAPI.getLogger(String.valueOf(Math.random()));
                    }
                }).start();
        } What is the expected output? What do you see instead? Expect the program to finish, doesn't always finish.  Non-synchronized puts/gets to HashMap can result in an infinite loop. What version of the product are you using? On what operating system? esapi-2.0.1.jar
$ java -version
java version "1.6.0_32"
Java(TM) SE Runtime Environment (build 1.6.0_32-b05)
Java HotSpot(TM) Server VM (build 20.7-b02, mixed mode) Please provide any additional information below. Should use Collections.synchronizedMap or ConcurrentHashMap

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=280_

kwwall:
Closed via PR #470

PR #470, opened 2019-01-13
JavaLogFactory Thread Safety #286

jeremiahjstacey:
Possible solution to Issue #286 to test and resolve the thread safety concerns of creating JavaLog references.

### DIFF
--- a/src/main/java/org/owasp/esapi/reference/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/reference/JavaLogFactory.java
@@ -49,15 +49,7 @@ public class JavaLogFactory implements LogFactory {
 	* {@inheritDoc}
 	*/
 	public Logger getLogger(Class clazz) {
-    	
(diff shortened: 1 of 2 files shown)


jeremiahjstacey:
A side effect from this change that should be called out:  

Previously This case would have been true:
```
Logger log1 = JavaLogFactory.getInstance().getLogger(Clazz.class);
Logger log2 = JavaLogFactory.getInstance().getLogger(Clazz.class.getName());
Assert.assertNotEquals(log1, log2)
```

Changes in this review will return the same logger in both cases.

kwwall:
Was that not the expected outcome though? At least if devs were creating
Logger instances via
        ESAPI.logger()
which almost certainly was implemented as a Singleton pattern.

Assuming that is the case, do you really think it is worth a specific
mention in the release notes?

-kevin
--

kwwall:
Closing issue #286

Issue #31, opened 2014-11-13
MySQL CODEC : "_" character not handled properly ?


meg23:
_From [Mikael.M...@gmail.com](https://code.google.com/u/112394004954344914018/) on July 24, 2009 05:41:04_

The "_" and" &#37;" characters are escaped but http://mirror.yandex.ru/mirrors/ftp.mysql.com/doc/refman/5.0/en/string-syntax.html specifies that these characters have specific behaviours :"If you use “&#37;”
or “\_” outside of pattern-matching contexts, they evaluate to the strings
“&#37;” and “_”, not to “&#37;” and “_”. " 

Example : if we have a table FOO with a row with field VALUE_FOO='TEST_1', 
the SQL command to retrieve this row : SELECT \* FROM FOO WHERE
VALUE_FOO=encodeForSQL('TEST_1') will become : SELECT \* FROM <table> WHERE
<field>='TEST_1' which will not return the row.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=21_

drmyersii:
Where has this thread left off? I'd like to start contributing, but I'm not sure what exactly needs fixed in this issue. Can someone with more domain knowledge elaborate a bit more? Thanks!

kwwall:
David,

I would start with a new Junit test case in CodecTest.java. Maybe something
along the lines that Mikael suggested in his assessment earlier post on
this topic:
-------
Example : if we have a table FOO with a row with field VALUE_FOO='TEST_1',
the SQL command to retrieve this row : SELECT * FROM FOO WHERE
VALUE_FOO=encodeForSQL('TEST_1') will become : SELECT * FROM

WHERE ='TEST_1' which will not return the row.
------
might make a good test case. (See comment earlier in this thread where this
is mentioned.) See if the new test works as expected.  If so, I guess it's
already fixed ( but leave the JUnit test as confirmation.)

It sounds as though it was supposedly fixed in the 1.4 release, but not
maybe not for 2.x, soI if it's broke for your test for 2.xand under the
"develop" branch, then look under the 1.4 Branch in GitHub and look at
either MySQLCodec.java or Codec.java. If it was fixed in 1.4, it would
probably be in one of those files (or perhaps in a config file like
ESAPI.properties).

Hope that helps enough to get your started.

-kevin
Sent from my Droid; please excuse typos.

drmyersii:
[...]

PR #472, opened 2019-01-21
Issue #31 MySQLCodec Updates

jeremiahjstacey:
### Summary
Work to make the MySQLCodec compliant with the OWASP recommended practices as defined in
https://www.owasp.org/index.php/SQL_Injection_Prevention_Cheat_Sheet#MySQL_Escaping

https://github.com/ESAPI/esapi-java-legacy/commit/5a6fa1e8a9448249756c90e0421d39f23f05ec1b

### Functional change in ANSI Mode:
  **Removed handling for double-quote escapes**

>  ANSI_QUOTES mode specifically says that double-quotes cannot be used to capture string literals. 
> 
>     With ANSI_QUOTES enabled, you cannot use double quotation marks to quote literal strings because they are interpreted as identifiers.
> 
> Later in the same document it states.
> 
>     "A ' inside a string quoted with " needs no special treatment and need not be doubled or escaped. In the same way, " inside a string quoted with ' needs no special treatment."
> 
> So, since ANSI_QUOTES only allows literals in  single ticks (which we escape), there should be no issue if those literals contains double quotes.  The handling of double quotes that we had (replacing with nothing) only appears to be valid if it is inside of a double-quoted string
> 
>     A " inside a string quoted with " may be written as "".
> 
> In ANSI_QUOTES mode, that's actually bad sql (from what I think I understand).
> 
> I pulled this information from https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

### Functional change in the deprecated Constructor 
Updated mode lookup to throw an exception as soon as the mode cannot be resolved as opposed to allowing the mode to be null and possibly throw NPE's at runtime.

https://github.com/ESAPI/esapi-java-legacy/commit/9ad908ae262ec48f2ce7d4240abdc2a5a06b9fb0

### All other commits are increasing basic test coverage.
----
### Issue #31 investigation summary
  Issue 31 originally targeted the context-based handling of percent (%) and underscore (_) based on where it appeared in the statement.  It was stated that the behavior was corrected on the ESAPI 1.4 branch.
Upon investigating the changes on the 1.4 branch I found that the resolution was to alter the default mode applied to the MySQLCodec from ANSI to STANDARD (In that version mode had a default value).  This is confirmed by later conversations in the issue.
In the current structure of the code there is no default value, and the mode configuration is required to establish an instance of the class.  I believed that the issue was effectively OBE'd by the change in the constructor logic of the class.
With that in mind I went about adding tests asserting ESAPI MySQLCodec compliance to the OWASP recommended strategies, which is where the changes above originated.  There were two logic cases where the modes did not perform to the recommendation, which have been addressed in this effort.

I believe this work resolves Issue #31 by showing that we are following industry recommendations for the necessary cases and that no additional logic or handling is integrated which would cause divergence from that intent.


jeremiahjstacey:
Hold on merge. Have a test impacted in the EncoderTest.  Working on it now.

jeremiahjstacey:
The test that tripped me up is EncoderTest  testMySQLANSIModeQuoteInjection

It is expecting that the double quote is escaped.  If we're in agreement that the double quotes don't need to be handled in this way then I believe I am going to remove the test.  All of the checks for the MySQLCodec have been moved into the test specific to the class.

jeremiahjstacey:
Changset attached to PR.  All tests passed, should be gtg.

kwwall:
@jeremiahjstacey wrote:
> To answer your question specifically, I can't fix this test. Based on what I think I understand this test is invalid against the intent of the implementation.

What I meant by "fix" and what _you_ mean by "fix" are 2 different things. What I mean was why not _reform_ the test so that it passes and then add a comment there about how double quotes do not require special handling for ANSI_QUOTE mode (aka, MySQLCodec.Mode.ANSI). You could just do that by changing the assertion, right? E.g., something like this:
```
        // In ANSI mode, double quotes need no special handling.
        assertEquals("MySQL Ansi Quote Injection Bug", "\" or 1=1 -- -", instance.encodeForSQL(c, "\" or 1=1 -- -"));
```
and then the test passes.

I think that's better than just removing the test, because someone is not going to come back and think "uh oh, an SQLi vulnerability is still present and they are trying to sneak it past us because they don't know how to fix it so they just simply removed the test and hoped no one noticed". You could argue that the test is unneeded and probably would be right, but _perception_ is just as important as reality in these cases. People act on their perception of truth rather than on actual truth so it's important IMO not to make it appear as we are trying to do something on the sly here.

jeremiahjstacey:
You make a very good point, thank you for calling that out..  I've reintroduced the test.  The comparison condition was updated to match the current impl's expectation and a comment was added to capture the intent.

PR #484, opened 2019-04-06
Misc prep issue483

kwwall:
Thanks to @jeremiahjstacey all tests pass now. Still working on the Dependency Check issues, but let's not let that hold us back. The worst case is we may need to update a few more dependencies.


jeremiahjstacey:
Reviewed changes, all looks good.  Merging.

PR #487, opened 2019-04-29
Master branch updates for ESAPI-2.2.0.0-RC2

kwwall:
Corresponds to signed tag 'esapi-2.2.0.0-RC2'.

### DIFF
--- a/CONTRIBUTING-TO-ESAPI.txt
+++ b/CONTRIBUTING-TO-ESAPI.txt
@@ -69,7 +69,8 @@ Steps to work with ESAPI:
        forked repo created in previous step)
     3. Create a new branch to work on an issue. I usually name the branch
        'issue-#' where '#' is the GitHub issue # is will be working on, but
-       you can call it whatever.
+       you can call it whatever. E.g.,
+            git checkout -b issue-#
     4. Work on the GitHub issue on this newly created issue-# branch.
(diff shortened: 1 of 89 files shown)


fransonsr:
When was ESAPI-2.2.0.0-RC2 released? Our applications fail to load `ESAPI.properties`. The change seems to be with https://github.com/xeno6696/esapi-java-legacy/commit/283753841aa927eb535ddfddb5c899fd7721d39d 
```
in = loaders[i].getResourceAsStream(DefaultSearchPath.ROOT.toString());
```
This change attempts to load "ROOT" and not `fileName` (aka., "ESAPI.properties"). We are picking up this RC version via the transitive dependency in `encoder-esapi`:
```
        <dependency>
            <version>[2.0,3)</version>
        </dependency>
```

fransonsr:
I am going to pin our app to esapi version 2.1.0.1 in the meantime.

xeno6696:
that sucks.  SHould be an easy fix.

xeno6696:
That bug will be fixed in the next RC.

xeno6696:
To answer your question @fransonsr it was literally just released today, so that’s some quick feedback :-).  That said, while it was recommended in the past to compile your own version with your own properties, I’ve only ever done that myself once.

fransonsr:
Thanks.

PR #489, opened 2019-05-10
loading from the root of the application was broken

JoergAdler:
Hi,

with 2.2.0-RC2 a esapi.properties in the root of the application wasn't loadable anymore. I've done a fix for this.

### DIFF
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -48,7 +48,7 @@ import org.owasp.esapi.errors.ConfigurationException;
  * 2) Inside the System.getProperty( "org.owasp.esapi.resources" ) directory.
  * You can set this on the java command line as follows (for example):
  * <pre>
- * 		java -Dorg.owasp.esapi.resources="C:\temp\resources"
+ *      java -Dorg.owasp.esapi.resources="C:\temp\resources"
  * </pre>
  * You may have to add this to the start-up script that starts your web server. For example, for Tomcat,
  * in the "catalina" script that starts Tomcat, you can set the JAVA_OPTS variable to the {@code -D} string above.
@@ -83,12 +83,12 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     }
     
     private Properties properties = null;
-    private String cipherXformFromESAPIProp = null;	// New in ESAPI 2.0
-    private String cipherXformCurrent = null;		// New in ESAPI 2.0
+    private String cipherXformFromESAPIProp = null; // New in ESAPI 2.0
+    private String cipherXformCurrent = null;       // New in ESAPI 2.0
 
-	/** The name of the ESAPI property file */
-	public static final String DEFAULT_RESOURCE_FILE = "ESAPI.properties";
-	
+    /** The name of the ESAPI property file */
+    public static final String DEFAULT_RESOURCE_FILE = "ESAPI.properties";
+    
     public static final String REMEMBER_TOKEN_DURATION = "Authenticator.RememberTokenDuration";
     public static final String IDLE_TIMEOUT_DURATION = "Authenticator.IdleTimeoutDuration";
     public static final String ABSOLUTE_TIMEOUT_DURATION = "Authenticator.AbsoluteTimeoutDuration";
@@ -98,7 +98,7 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     public static final String MAX_OLD_PASSWORD_HASHES = "Authenticator.MaxOldPasswordHashes";
 
     public static final String ALLOW_MULTIPLE_ENCODING = "Encoder.AllowMultipleEncoding";
-    public static final String ALLOW_MIXED_ENCODING	= "Encoder.AllowMixedEncoding";
+    public static final String ALLOW_MIXED_ENCODING = "Encoder.AllowMixedEncoding";
     public static final String CANONICALIZATION_CODECS = "Encoder.DefaultCodecList";
 
     public static final String DISABLE_INTRUSION_DETECTION  = "IntrusionDetector.Disable";
@@ -113,9 +113,9 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     public static final String RANDOM_ALGORITHM = "Encryptor.RandomAlgorithm";
     public static final String DIGITAL_SIGNATURE_ALGORITHM = "Encryptor.DigitalSignatureAlgorithm";
     public static final String DIGITAL_SIGNATURE_KEY_LENGTH = "Encryptor.DigitalSignatureKeyLength";
-    			// ==================================//
-    			//		New in ESAPI Java 2.0		 //
-    			// ================================= //
+                // ==================================//
+                //      New in ESAPI Java 2.0        //
+                // ================================= //
     public static final String PREFERRED_JCE_PROVIDER = "Encryptor.PreferredJCEProvider";
     public static final String CIPHER_TRANSFORMATION_IMPLEMENTATION = "Encryptor.CipherTransformation";
     public static final String CIPHERTEXT_USE_MAC = "Encryptor.CipherText.useMAC";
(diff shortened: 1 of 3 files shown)


JoergAdler:
Sorry for the whitespace changes. It seems, that my eclipse is doing weird things :-(

src/test/java/org/owasp/esapi/reference/DefaultSecurityConfigurationTest.java
```diff
@@ -18,467 +18,475 @@
...
-		assertNull(secConf.getValidationPattern("Test2"));
-		assertNull(secConf.getValidationPattern("TestC"));
-		
-		secConf = new DefaultSecurityConfiguration("ESAPI-DualValidatorFileChecker.properties");
-		assertEquals(patternOrNull(secConf.getValidationPattern("Test1")), "ValueFromFile1");
-		assertEquals(patternOrNull(secConf.getValidationPattern("Test2")), "ValueFromFile2");
-		assertNull(secConf.getValidationPattern("TestC"));
+    @Test
+    public void testRootCPLoading(){
+        DefaultSecurityConfiguration secConf = new DefaultSecurityConfiguration("ESAPI-root-cp.properties");
```
kwwall: @JoergAdler Rather creating yet another variant of **ESAPI.properties** under "**src/test/resources/esapi**", could you please change your test to use one of these:

```
src/test/resources/esapi/ESAPI-DualValidatorFileChecker.properties
src/test/resources/esapi/ESAPI.properties
src/test/resources/esapi/ESAPI-SingleValidatorFileChecker.properties
src/test/resources/esapi/ESAPI-QuotedValidatorFileChecker.properties
src/test/resources/esapi/ESAPI-test-2.properties
src/test/resources/esapi/ESAPI-CommaValidatorFileChecker.properties
src/test/resources/esapi/ESAPI-test.properties
```
instead? There are already way too many of these and every time we add a new property to ESAPI or delete an old property, I feel compelled to keep all of these others in sync with the main one in `configuration/esapi/ESAPI.properties`, (and admittedly, have been doing a poor job) in part because we have been unable to figure out how to get the stuff under 'configuration/' into the 'sources' jar so people keep grabbing on of the test ones from `src/est/resources/esapi` and using those for production. (Okay; we do know how to get it into the sources jar; we just haven't figured out how to do that with also getting it built into the regular **esapi-**_release_**.jar**.)

I am fine with adding a custom property or two to one of these others. I'm _not_ fine with creating yet another **ESAPI.properties** file under `src/test/resources/esapi`.

src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
```diff
@@ -1381,29 +1381,29 @@ public String getStringProp(String propertyName) throws ConfigurationException {
...
-    		this.path = s;
-    	}
-    	
-    	public String value(){
-    		return path;
-    	}
+        
+        RESOURCE_DIRECTORY("resourceDirectory/"),
+        SRC_MAIN_RESOURCES("src/main/resources/"),
+        ROOT(""),
```
kwwall: @xeno6696 Can you confirm that this is the correct value?

xeno6696: Doesn’t look right.  I would have just restored the value I clobbered in the if block

src/test/java/org/owasp/esapi/reference/DefaultSecurityConfigurationTest.java
JoergAdler: Hi. 

The point of this test is to make sure, that an esapi-properties from the root of the classpath (cp) is loaded. 
So it has to be in src/test/resources/ and has to have a different name than the other ones, to prove, if not one of the other files is loaded accidentally.

src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
JoergAdler: Unfortunately it is the right value. As you can read here: 
http://www.thinkplexx.com/learn/howto/java/system/java-resource-loading-explained-absolute-and-relative-names-difference-between-classloader-and-class-resource-loading

"There are ClassLoader.getResource() and Class.getResource() and they work differently [...] The name should not have a leading “/”."

Please also check the new added test in DefaultSecurityConfigurationTest.

src/test/java/org/owasp/esapi/reference/DefaultSecurityConfigurationTest.java
jeremiahjstacey: I was looking at this a little bit, and I don't see a better way to prove the case being targeted by this change.   I will try to do a little more independent verification on it, but I think this change is a good thing.

@kwwall  if we could ensure that all copies of the ESAPI.properties in test scope emulate the values in the src/main directory would that alleviate some concern?

There may be some magic we can do with maven resource copy to duplicate the ESAPI.properties file into test scope with new names, and possibly new values for certain keys.  It's been a little while since I looked at it, but I believe this is a feasible thing.

If we did that then the files would stay in-sync with the primary properties.  In addition, only the src/main version would be in source control.  All others would be created during mvn package for the purpose of unit testing. We may even be able to explicitly add all src/test/**/ESAPI*.properties to the .gitignore file at that point so they don't get committed back up.

Aside from changing how the files are managed or updating the specific packaging of the resources in the delivery I'm not sure there's another way to address it.  Both of those efforts feel larger than this review though, so I hesitate to say that this change requires either of those things to be done first.  Especially when this is consistent with the current pattern and we are aware of the issue and are working to address it.

---
The one thing that may work (which I haven't tested) would be put this file with the other test resource files.  In the test use the beforeAll to copy the resource content to the root of the classpath; likewise, the afterall must remove the created file.
This feels heavy-handed to me, but may be another feasible interim solution while the larger resource management problems are being addressed.

src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
xeno6696: @JoergAdler Sorry for the delay.  My concern is that what you did here, isn't what was done in the past:

```
					// try root directory path given by user
					if (in == null){
						in = loaders[i].getResourceAsStream(fileName);
					}
					
```
Which looked erroneous to me, but it IS what we _were_ doing.  I think if we at least re-include that case I'm fine with the rest of the changes.  I'm still not sure why that case was referred to as "root."  I wanted to debug the old 2.1.0.1 release but that would have required me getting an older JDK and I didn't care all that much.

src/test/java/org/owasp/esapi/reference/DefaultSecurityConfigurationTest.java
kwwall: Certainly we could less the possible confusion (and my objection!) if we strip the new 'ESAPI-root-cp.properties' to the bare bones properties of what is actually needed here. I'm pretty sure that we don't need anything BUT
`ESAPI.Logger=org.owasp.esapi.reference.Log4JLogFactory`
and maybe the rest of the 'Logger.*' property name / value pairs. But other than the ESAPI logger properties, I don't think anything else should be required. If @JoergAdler agrees to do that, then I suppose I can live with the results.

kwwall:
It has been too long and the author of this PR has not responded. I am still going to include this fix in the 2.2.0.0 release, but I can wait no longer for the PR to get cleaned up (all the white-space diffs, plus the new property file really only needs a _single_ property in it), so I am going to reject this PR. I will still credit @JoergAdler in the 2.2.0.0 release notes though as contributing toward the fix. Thanks Jörg.

JoergAdler:
Hi, 

Sorry that I not responded for a long time. 

When I get it right, the following problems still exists in the pull request: formatting and the property file should have only one property. 

If you want, I could fix still fix this and reopen the pull request?

kwwall:
No worries. I've already made the changes based on your changes and tests, added some comments and additional tests, reduced the ESAPI-root-cp.properties file to a single property (and, it turns we really didn't need it to test this; using
new DefaultSecurityConfiguration("esapi/SingleValidatorFileChecker.properties")
or something similar works as well. But I kept your file and just added a second test for something like that latter. I will mention you in our release notes though.

Oh, one last thing. @xeno6696 mentioned in https://github.com/ESAPI/esapi-java-legacy/pull/489#discussion_r286674928 that:
`			in = loaders[i].getResourceAsStream(fileName);`
Since the enum ROOT now has a value of an empty string, that is in fact now logically equivalent to your fix of:
`                        in = loaders[i].getResourceAsStream(DefaultSearchPath.ROOT.value() + fileName);`

So that makes we feel a lot better about all this. But like I said, I already extracted your changes from your PR and pulled them into my commits, so don't sweat it.

I'm pretty much done with everything other than closing out a couple of more issues and updating the release notes, but I'm hoping that I should have the official 2.2.0.0 release out sometime late Sunday evening if all goes to plan.

PR #490, opened 2019-05-11
enhance: cache class and method to avoid reading each time

hellyguo:
**Before enhance**:
> Benchmark                        Mode  Cnt       Score       Error  Units
> ObjFactoryBenchmark.benchmark1  thrpt   30  337882.228 ±  3149.328  ops/s
> ObjFactoryBenchmark.benchmark2  thrpt   30  279072.632 ± 17234.329  ops/s

**After enhanced**:
> Benchmark                        Mode  Cnt       Score       Error  Units
> ObjFactoryBenchmark.benchmark1  thrpt   30  377886.687 ± 13520.653  ops/s
> ObjFactoryBenchmark.benchmark2  thrpt   30  358412.164 ± 14730.155  ops/s

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -381,6 +381,16 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.21</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.21</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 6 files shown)


kwwall:
@hellyguo First, thanks for doing this. In general, I think it out to help. This has been something on my TODO list for a few years, but with all other higher priority things, I just never found the time to do this. (I thought of doing this when I originally created ObjFactory but I was just new to ESAPI at the time and I wasn't even sure how the others would accept my refactoring things to use ObjFactory at all so I thought it best to keep it initially simple.)

Anyhow, I've already went through some of the code. I will do a more detailed code review on this shortly, but in the meantime had two general comments that I wanted to give your.

1. In the pom.xml, the 2 org.openjdk.jmh jars should be added as 'test' scope. Otherwise, I will reject that change.
2. Please undo the messing with the imports, including the ordering. Deleting used imports is fine, but in particular, we try to avoid wildcards in importing classes from packages such as the 'org.owasp.esapi.crypto.*' that you changed in JavaEncryptor.java for example. There is a reason for the expansion is it allows us to do recursive greps to succinctly look for where classes are used. I do that quite often for reasons that I won't go into (but mostly has to do with me not generally using an IDE; I'm old school and use gvim for most editing tasks).

Once you address those 2 things, I will start with a more detailed code review.

Thanks again for doing this though.

hellyguo:
Sorry for those 2 things. I will close this pull request. And I will create new one after I clean the code.

BTW. I love GVIM too.

PR #497, opened 2019-06-10
Prepare 2.2.0.0-RC3

kwwall:
Preparation for release candidate 2.2.0.0-RC3


src/test/resources/esapi/ESAPI-CommaValidatorFileChecker.properties
```diff
@@ -1,5 +1,35 @@
...
+#   should retrieve THAT version from the official GitHub report from
+#       https://github.com/ESAPI/esapi-java-legacy/blob/develop/configuration/esapi/ESAPI.properties
+#   but make sure that you select it from the 'master' branch (which will
+#   correspond to the latest official ESAPI relase). Sorry for the
+#   inconvenience. We are trying to figure out how to get it to the official
+#   "esapi-<releaseVersion>-sources.jar fiel available from Maven Central, but
+#   in the meantime, you will have to get it from GitHub.
+#
+#   PLEASE do not base your production use of ESAPI on this TEST version of
+#   ESAPI.properties as this test version has been dummed down in several places
```
xeno6696: Dummed?  was that on purpose!  :-D  If it was it's genius!

src/main/java/org/owasp/esapi/util/ObjFactory.java
```diff
@@ -30,11 +30,11 @@
  * 		...
  * 		// Typically these would be populated from some Java properties file
  * 		String barName = "com.example.foo.Bar";
- * 		String beerBrand = "com.example.brewery.Guiness";
+ * 		String beerBrand = "com.example.brewery.Guinness";
```
xeno6696: You are cracking me up this morning!

src/main/java/org/owasp/esapi/util/ObjFactory.java
```diff
@@ -45,10 +45,11 @@
  */
 public class ObjFactory {
 
-	private static final int CACHE_INITIAL_CAPACITY = 4096;
+    private static final int CACHE_INITIAL_CAPACITY = 32;
```
xeno6696: Any real motivation for this change?  are these bytes or MB?

xeno6696:
No issues.  Merge is Good.  Concur @jeremiahjstacey ?

src/main/java/org/owasp/esapi/util/ObjFactory.java
jeremiahjstacey: If I'm reading the docs right, this is just the initial quantity of elements that the Map can hold.  When 'load factor' % of 'initial capacity' is reached, the map will increase in size.  

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#ConcurrentHashMap-int-float-

jeremiahjstacey:
Looks good to me @xeno6696 !

src/test/resources/esapi/ESAPI-CommaValidatorFileChecker.properties
kwwall: While I'd love to claim credit for the incredible irony, I'll have to blame it on too many 'Guiness'. ;-) Speaking of which, I can't believe I misspelled that and never noticed. (Okay, I think I maybe had fixed it in the private BitBucket repo.)

src/main/java/org/owasp/esapi/util/ObjFactory.java
kwwall: @xeno6696 So, misspelled, and apparently ignored until I finally fix it. Uh huh. You guys are missing all my good jokes. Did you see the one on line 111 of ObjFactoryTest.java. That's been there for years as well and no one has said anything...probably because everyone agrees. :)

src/main/java/org/owasp/esapi/util/ObjFactory.java
kwwall: @jeremiahjstacey Exactly. And no way we need 4K. There are only 11 calls to `ObjFactory.make()` in ESAPI.java which will be the usual case, and it does automatically grow as needed, so there's little point in going crazy here. Even 32 is probably more than will ever be needed.

kwwall:
@xeno6696 @jeremiahjstacey Maybe I should quote 'dumming'. ;-)

PR #501, opened 2019-06-24
Final preparation for ESAPI 2.2.0.0 release

### LINKED ISSUE #488: Missed a legal input case in DefaultSecurityConfiguration.java
When was ESAPI-2.2.0.0-RC2 released? Our applications fails to load `ESAPI.properties`. The change seems to be with https://github.com/xeno6696/esapi-java-legacy/commit/283753841aa927eb535ddfddb5c899fd7721d39d 
```
in = loaders[i].getResourceAsStream(DefaultSearchPath.ROOT.toString());
```
This change attempts to load "ROOT" and not `fileName` (aka., "ESAPI.properties"). We are picking up this RC version via the transitive dependency in `encoder-esapi`:
```
        <dependency>
            <version>[2.0,3)</version>
        </dependency>
```

_Originally posted by @fransonsr in https://github.com/ESAPI/esapi-java-legacy/pull/487#issuecomment-487708278_

### LINKED ISSUE #499: ValidatorTest.isValidDirectoryPath() has tests that fail under Windows if ESAPI tests run from different drive where Windows installed
@davewichers reported this issue.

On Dave's Windows box, ESAPI is installed under D: drive and Windows is installed on C: drive. This results in several failed tests in ValidatorTest.isValidDirectoryPath().

These tests fail because when `DefaultValidator.isValidDirectoryPath` is called with 'input' and 'parent', one of the tests verification checks that takes place is to make sure that the specified 'input' directory is "inside" the specified 'parent' directory. But for Dave's case, running on D: drive, 'parent' (which is set to '`new File("/");`') will end up having a canonical path of "D:\\", so in DefaultValidator, the check results in
`"C:\\".startsWith("D:\\")`
which obviously fails.

### LINKED ISSUE #500: Suppress noise from ESAPI searching for properties and stop ignoring important IOExceptions
There were related changes between
```
src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoaderFactory.java
src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
```

that together should result in suppressing some of the unnecessary noise when the `System` property '**org.owasp.esapi.logSpecial.discard**' is set to '**true**'. Also important `IOExceptions` that were simply being ignored before (after a error message to stderr and/or stdout) is no longer being ignored.

kwwall:
Close #488 
Close #499 
Close #500 
See 2.2.0.0 release notes in documentation/esapi4java-core-2.2.0.0-release-notes.txt

Do **NOT** merge until @davewichers has a chance to verify that issue #499 is actually fixed on Windows.


jeremiahjstacey:
Preconditions met.  #499 has been confirmed as resolved by @davewichers.

Issue #256, opened 2014-11-13
Whitespace in JavaEncryptor


meg23:
_From [noloa...@gmail.com](https://code.google.com/u/114558122492435650190/) on October 15, 2011 18:08:08_

Whitespace

**Attachment:** [JavaEncryptor.java.patch](http://code.google.com/p/owasp-esapi-java/issues/detail?id=249)

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=249_

kwwall:
Confirmed that this patch is only cleanup of whitespace so this is indeed low priority. (In the mean time, set your tabsize to 4 if it bothers you.)

jeremiahjstacey:
Closing with the merge of PR #505

PR #505, opened 2019-07-07
Close issue #256. White-space clean up.

kwwall:
Close GitHub issue #256 
This looks  like a lot of diffs, but they are all white-space related diffs.
Changed all tabs to the equivalent of spaces, with tabsize set to 4 spaces.
'git diff -w' should should minimal changes. (I think there was few places where tab before '//' comment was changed to spaces, which might show up as a difference.)


jeremiahjstacey:
Verified incoming content contains only whitespace modifications

PR #508, opened 2019-08-02
Resolves #226 - Corrected docs for the  bounded, numeric, random methods

### LINKED ISSUE #226: Javadoc Inaccuracy in getRandomInteger() and getRandomReal()
_From [chris.fr...@gmail.com](https://code.google.com/u/117338994236964423833/) on March 27, 2011 18:31:26_

What steps will reproduce the problem? 1.Call ESAPI.randomizer().getRandomInteger(1, 2) multiple times

What is the expected output? 
According to http://owasp-esapi-java.googlecode.com/svn/trunk_doc/latest/org/owasp/esapi/reference/DefaultRandomizer.html#getRandomInteger(int , int), the expected output is the random return of sometimes 1 (min) and sometimes 2 (max).

What do you see instead?
I always get 1 in return. What version of the product are you using? On what operating system? rc4, but the latest source code on the repository contains the same discrepancy vs. the javadoc Does this issue affect only a specified browser or set of browsers? no. Please provide any additional information below. ---------

On http://owasp-esapi-java.googlecode.com/svn/trunk_doc/latest/org/owasp/esapi/reference/DefaultRandomizer.html#getRandomInteger(int , int)

we read, among other things:

```
Parameters:
min - the minimum integer that will be returned
max - the maximum integer that will be returned
```

I believe that "max" is larger than the maximum integer that will be returned. In other words, a call to ESAPI.randomizer().getRandomInteger(1, 2) will never return anything else than 1.

A formulation like this would be more accurate (taken from http://download.oracle.com/javase/6/docs/api/java/util/Random.html#nextInt(int)) :

```
Returns a pseudorandom, uniformly distributed int value between min (inclusive) and max (exclusive)
```

The ESAPI DefaultRandomizer uses the above mentioned JavaSE6 Random method (inherited by SecureRandom)

```
 public int getRandomInteger(int min, int max) {
        return secureRandom.nextInt(max - min) + min;
}
```

I see however that the javadoc for the ESAPI Interface http://owasp-esapi-java.googlecode.com/svn/trunk_doc/latest/org/owasp/esapi/Randomizer.html#getRandomInteger(int , int) contains the same definition as does the DefaultRandomizer. In case the DefaultRandomizer is implementing the Interface incorrectly, then the reference implementation would need to be changed to meet the javadoc in stead. If I understand correctly, however, it is more conventional for the returned values of such methods to exclude the parameters' "max" value, rather than to include it.

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=217_

ZiluckMichael:
Added the documentation for the method referenced in #226
Also added documentation for getRandomReal(float,float) so it is not left ambiguous

Although the original issue does not mention getRandomReal, I thought it would be appropriate to add the documentation there as well for additional clarity. As a developer, seeing the documentation on getRandomInteger could cause confusion about the behavior of getRandomReal.

### DIFF
--- a/src/main/java/org/owasp/esapi/Randomizer.java
+++ b/src/main/java/org/owasp/esapi/Randomizer.java
@@ -1,15 +1,15 @@
 /**
  * OWASP Enterprise Security API (ESAPI)
- * 
+ *
  * This file is part of the Open Web Application Security Project (OWASP)
  * Enterprise Security API (ESAPI) project. For details, please see
  * <a href="http://www.owasp.org/index.php/ESAPI">http://www.owasp.org/index.php/ESAPI</a>.
  *
  * Copyright (c) 2007 - The OWASP Foundation
- * 
+ *
  * The ESAPI is published by OWASP under the BSD license. You should read and accept the
  * LICENSE before you use, modify, and/or redistribute this software.
- * 
+ *
  * @author Jeff Williams <a href="http://www.aspectsecurity.com">Aspect Security</a>
  * @created 2007
  */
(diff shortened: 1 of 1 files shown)


jeremiahjstacey:
From what I'm seeing the resolution is inline with the adjustments requested in the issue.   I believe this is good to merge.  @kwwall I can merge after you get a chance to review (or feel free to merge if you agree and have capacity)

kwwall:
@jeremiahjstacey Looks good to me. Go ahead and do the merge and close GitHub issue #226

Issue #515, opened 2019-09-13
NullPointerException can result from line 163 of SecurityWrapperRequest.java:


mvkendall:
Cookie n = new Cookie(name, value); // when name is null NPE is thrown

java.lang.NullPointerException
        at javax.servlet.http.Cookie.isToken(Cookie.java:514)
        at javax.servlet.http.Cookie.<init>(Cookie.java:135)

xeno6696:
Would be lots faster if you'd fork the code and then add the null check, then submit a PR!  We need help on this project!

kwwall:
The problem I have with this is what is the best way to handle this? Any exception that is thrown will have to be an _unchecked_ exception because we can't change public interfaces without potentially breaking code and the default `NullPointerException` that is thrown is probably the a reasonable unchecked exception type here, although not the best choice. Since ESAPI is not setting the cookies in this case, but rather just grabbing the Cookie: HTTP request header from the HTTP request, it's not really appropriate (IMO at least) to throw an `IllegalArgumentException` here.

Now this case definitely _could_ happen from an attack, so yes, we should to do something here. IMO throwing any sort of `Error` for an unchecked exception is just wrong, so it will have to be some sort of `RuntimeException`. ESAPI has 3 such exceptions: the base class `EnterpriseSecurityRuntimeException` and subclasses `EncryptionException` and `IntrusionException`. Unless one of you can think of how this might accidentally (and _commonly_) arise from a programming error, I think we should go with `IntrusionException`. As per the Javadoc for `IntrusionException`:

> An `IntrusionException` should be thrown anytime an error condition arises that is likely to be the result of an attack in progress.

Specifically, I think we should use the 3 arg CTOR here. I think that the 'logMessage' argument should be something like "HTTP request detected Cookie: header where cookie did not have a required name". (I don't think we should log the cookie _value_ because it potentially could be a sensitive value like a session ID, etc. that really ought not to be logged.) As for the 'userMessage', I'm open to suggestions there. And for the 'cause' argument, just pass in the NPE that is caught. But that's my $.02 on this if @mvkendall wants to do a PR for us.

Interested in hearing other's opinions about this though.

kwwall:
I took another closer look at this. Seems to me that line 157, which is this:

`                String name = ESAPI.validator().getValidInput("Cookie name: " + c.getName(), c.getName(), "HTTPCookieName", sc.getIntProp("HttpUtilities.MaxHeaderNameSize"), true);`

[...]

PR #516, opened 2019-09-15
Issue 515

jeremiahjstacey:
Added tests for SecurityWrapperRequest.getCookies and updated the allowNull flag for the cookie name to false.


kwwall:
Close issue #515

PR #519, opened 2019-09-29
Issue 494 CSSCodec RGB Triplets

jeremiahjstacey:
@xeno6696  I think this achieves the capability you were looking for with Issue 494.  I have not worked much with the expectations of the Codec API so please let me know if I'm missing something.

As an aside, I did try to build the regex replacement capability as a separate class instance.  In future cases like this we may just need to figure out the regex and inject this piece into the process.

The new test cases pass, but I am currently seeing an issue with ValidatorTest.testIsValidSafeHTML and ValidatorTest.testGetValidSafeHTML.  I've verified that I am seeing the same failures across my fork of the develop branch.

### DIFF
--- a/src/main/java/org/owasp/esapi/codecs/CSSCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/CSSCodec.java
@@ -15,6 +15,10 @@
  */
 package org.owasp.esapi.codecs;
 
+import java.util.regex.Pattern;
+
+import org.owasp.esapi.codecs.ref.EncodingPatternPreservation;
+
 /**
  * Implementation of the Codec interface for backslash encoding used in CSS.
  * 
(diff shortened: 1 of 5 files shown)


src/main/java/org/owasp/esapi/codecs/CSSCodec.java
```diff
@@ -63,13 +80,15 @@ public Character decodeCharacter(PushbackSequence<Character> input)
...
 		Character first = input.next();
+		System.out.println("First: " + first);
 		if (first == null || first != '\\')
 		{
 			input.reset();
 			return null;
 		}
 
 		Character second = input.next();
+		System.out.println("Second: " );
```
jeremiahjstacey: This slipped in on me.  Getting it cleaned up..

jeremiahjstacey: https://github.com/ESAPI/esapi-java-legacy/pull/519/commits/377d377dcd68ccefa0841ac12e980729945363d9  Cleans up the debug content

Issue #143, opened 2014-11-13
Enchance encodeForOS to auto-detect the underling OS


meg23:
_From [manico.james@gmail.com](https://code.google.com/u/manico.james@gmail.com/) on July 26, 2010 16:48:03_

The encodeForOS function in the ESAPI for Java reference implementation should autodetect the OS, rather than require the caller to send in the type of OS to make it safe for. It should use the right codec automatically given the underlying OS, or throw some kind of notimplementedexception if it determines it’s on an OS that it doesn’t recognize (have an encoder for).

_Original issue: http://code.google.com/p/owasp-esapi-java/issues/detail?id=133_

xeno6696:
Hmm.  Can anyone retrieve Manico's patch?  @kwwall ?

kwwall:
@xeno6696 Yeah, I have all the patches from Google Code issues that were had corresponding GitHub issues that still open as of Feb 1, 2016. I will email all of them to you in a gzipped tarball.

jeremiahjstacey:
@xeno6696 @kwwall  I think this should be closed.  The issue is concerned with DefaultEncoder.encodeforOS.    It appears that method has been extended from taking only a String to taking the Codec and the String.  The only call in the baseline (non-test) is from the DefaultExecutor.  Looking at the constructor for the DefaultExecutor, it is already handling windows and unix systems:
```

 /**
     * Instantiate a new Executor
     */
    private DefaultExecutor() {
		if ( System.getProperty("os.name").indexOf("Windows") != -1 ) {
			logger.warning( Logger.SECURITY_SUCCESS, "Using WindowsCodec for Executor. If this is not running on Windows this could allow injection" );
			codec = new WindowsCodec();
		} else {
			logger.warning( Logger.SECURITY_SUCCESS, "Using UnixCodec for Executor. If this is not running on Unix this could allow injection" );
			codec = new UnixCodec();
		}
    }
```

Unless I'm misunderstanding, this issue is resolved by the strategy pattern codec from the DefaultExecutor and there is no additional work needed.

xeno6696:
I’m inclined to agree.  But maybe there should be a Linux check?

kwwall:
@xeno6696 The way that the code is written, it doesn't need a Linux check. The UmixCodec will work for Linux and the check is such that, if the OS is not Windows something or other, then the UnixCodec is used.

kwwall:
[...]

PR #520, opened 2019-10-04
OS Name DefaultExecutorTests #143

jeremiahjstacey:
Creating two tests to assert that the constructor of the DefaultExector
interprets the os.name system property.  If the name contains 'Windows'
the underlying Codec is a WindowsCodec.  If the literal 'Windows' is not
present, then the UnixCodec is applied to the instance.


xeno6696:
Can't really argue with improving test coverage!

PR #529, opened 2019-11-28
Esapi#527

ricksouza:
This new pull request makes the app and user info optional in ESAPI logger.

### DIFF
--- a/configuration/esapi/ESAPI.properties
+++ b/configuration/esapi/ESAPI.properties
@@ -391,6 +391,10 @@ Logger.LogServerIP=true
 Logger.LogFileName=ESAPI_logging_file
 # MaxLogFileSize, the max size (in bytes) of a single log file before it cuts over to a new one (default is 10,000,000)
 Logger.MaxLogFileSize=10000000
+# Determines wheter ESAPI should log the user info.
+Logger.UserInfo=true
+# Determines wheter ESAPI should log the app info.
+Logger.AppInfo=true
 
 
 #===========================================================================
--- a/src/main/java/org/owasp/esapi/SecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/SecurityConfiguration.java
@@ -730,6 +730,20 @@ public interface SecurityConfiguration extends EsapiPropertyLoader {
 	 */
 	@Deprecated
 	boolean getLogServerIP();
+	
+	/**
+	 * Returns whether ESAPI should log the user info. 
+	 * 
+	 * @return True if ESAPI should log the user info, False otherwise
+	 */
+	boolean getUserInfo();
+	
+	/**
+	 * Returns whether ESAPI should log the app info. 
+	 * 
+	 * @return True if ESAPI should log the app info, False otherwise
+	 */
+	boolean getAppInfo();
 
 	/**
 	 * Returns the current log level.
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -1077,6 +1077,20 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     public boolean getLogServerIP() {
     	return getESAPIProperty( LOG_SERVER_IP, true );
 	}
+    
+    /**
+   	 * {@inheritDoc}
+   	 */
+       public boolean getUserInfo() {
+       	return getESAPIProperty( LOG_USER_INFO, true );
+   	}
+       
+       /**
+   	 * {@inheritDoc}
+   	 */
+       public boolean getAppInfo() {
+       	return getESAPIProperty( LOG_APP_INFO, true );
+   	}
 
     /**
 	 * {@inheritDoc}
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -152,6 +152,8 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     public static final String LOG_ENCODING_REQUIRED = "Logger.LogEncodingRequired";
     public static final String LOG_APPLICATION_NAME = "Logger.LogApplicationName";
     public static final String LOG_SERVER_IP = "Logger.LogServerIP";
+    public static final String LOG_USER_INFO = "Logger.UserInfo";
+    public static final String LOG_APP_INFO = "Logger.AppInfo";
     public static final String VALIDATION_PROPERTIES = "Validator.ConfigurationFile";
     public static final String VALIDATION_PROPERTIES_MULTIVALUED = "Validator.ConfigurationFile.MultiValued";
     public static final String ACCEPT_LENIENT_DATES = "Validator.AcceptLenientDates";
--- a/src/main/java/org/owasp/esapi/reference/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/reference/JavaLogFactory.java
@@ -123,6 +123,12 @@ public class JavaLogFactory implements LogFactory {
     	/** Log the server ip? */
     	private static boolean logServerIP = ESAPI.securityConfiguration().getLogServerIP();
     	
+    	/** Log the user info */
+    	private static boolean logUserInfo = ESAPI.securityConfiguration().getUserInfo();
+
+    	/** Log the app info */ 
+    	private static boolean logAppInfo = ESAPI.securityConfiguration().getAppInfo();
+    	
         /**
          * Public constructor should only ever be called via the appropriate LogFactory
          * 
@@ -298,23 +304,25 @@ public class JavaLogFactory implements LogFactory {
             }
 
 			// log server, port, app name, module name -- server:80/app/module
-			StringBuilder appInfo = new StringBuilder();
-			if ( ESAPI.currentRequest() != null && logServerIP ) {
-				appInfo.append( ESAPI.currentRequest().getLocalAddr() + ":" + ESAPI.currentRequest().getLocalPort() );
-			}
-			if ( logAppName ) {
-				appInfo.append( "/" + applicationName );
-			}
-			appInfo.append( "/"  + moduleName );
-			
 			//get the type text if it exists
 			String typeInfo = "";
 			if (type != null) {
-				typeInfo += type + " ";
+				typeInfo += type;
 			}
 			
+			String userAndappInfo = "";
+			if(logUserInfo && logAppInfo)
+				userAndappInfo = getUserInfo() + " -> " + getAppInfo();
+			else if(logUserInfo)
+				userAndappInfo = getUserInfo();
+			else if(logAppInfo)
+				userAndappInfo = getAppInfo().toString();
+			
+			if(!userAndappInfo.isEmpty())
+				typeInfo += " ";
+			
 			// log the message
-			jlogger.log(level, "[" + typeInfo + getUserInfo() + " -> " + appInfo + "] " + clean, throwable);
+			jlogger.log(level, "[" + typeInfo + userAndappInfo + "] " + clean, throwable);
         }
 
         /**
@@ -358,6 +366,18 @@ public class JavaLogFactory implements LogFactory {
         public boolean isWarningEnabled() {
     	    return jlogger.isLoggable(Level.WARNING);
         }
+        
+        private StringBuilder getAppInfo() {
+			StringBuilder appInfo = new StringBuilder();
+			if ( ESAPI.currentRequest() != null && logServerIP ) {
+				appInfo.append( ESAPI.currentRequest().getLocalAddr() + ":" + ESAPI.currentRequest().getLocalPort() );
+			}
+			if ( logAppName ) {
+				appInfo.append( "/" + applicationName );
+			}
+			appInfo.append( "/"  + moduleName );
+			return appInfo;
+		}
         
         public String getUserInfo() {
             // create a random session number for the user to represent the user's 'session', if it doesn't exist already
--- a/src/main/java/org/owasp/esapi/reference/Log4JLogger.java
+++ b/src/main/java/org/owasp/esapi/reference/Log4JLogger.java
@@ -48,6 +48,12 @@ public class Log4JLogger extends org.apache.log4j.Logger implements org.owasp.es
 	
 	/** Log the server ip? */
 	private static boolean logServerIP = ESAPI.securityConfiguration().getLogServerIP();
+	
+	/** Log the user info */ 
+	private static boolean logUserInfo = ESAPI.securityConfiguration().getBooleanProp(DefaultSecurityConfiguration.LOG_USER_INFO);
+
+	/** Log the app info */ 
+	private static boolean logAppInfo = ESAPI.securityConfiguration().getBooleanProp(DefaultSecurityConfiguration.LOG_APP_INFO);
 
 	public Log4JLogger(String name) {
 		super(name);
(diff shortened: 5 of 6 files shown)


jeremiahjstacey:
Unless I'm mistaken, if you update to the newer SLF4JLogger in the latest ESAPI release you will not receive any of the user related verbosity.

Both the Log4JLogger and the JavaLogger (inside JavaLogFactory) use a delegate method that appends the user data.  The SLF4J implementation only acts as a passthrough providing newline protection and HTML Encoding.  HTML Encoding only occurs if the property `Logger.LogEncodingRequired` is set to `true`


- https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogger.java
- https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/reference/Log4JLogger.java
- https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/reference/JavaLogFactory.java

Should just need to set `ESAPI.Logger=org.owasp.esapi.logging.slf4j.Slf4JLogFactory` in your ESAPI.properties file.

Using the newer logger may avert the need for more configuration in the project.

cristiantm:
> Using the newer logger may avert the need for more configuration in the project.

However this does not apply for projects that are Lo4j/Java Logger based (our case). 

Regarding the PR, it covers backward compatibility with ESAPI.properties by using a default to true value. So users without this setting will notice no change.

xeno6696:
@cristiantm I don't necessarily disagree, but from a security perspective, log4j1 is well, beyond deprecated.  It's dead.  It hasn't been updated in 7 years, because OF log4j2 and slf4j.  As a security library I'm not comfortable making it easier for organizations to maintain backwards compatability for insecure libraries.  @kwwall and I have been making conscious decisions to eliminate support for abandoned platforms.  We made an agreement not to support Java versions older than 2 major releases back, and while we still support back to at least 1.7, THAT support is slated to be killed.  

I'm disinclined to accept this PR because I'd rather have organizations relying on log4j1 refactor their code to a newer, more secure library.  I'll allow my colleagues to have their say, but my vote here is no.  [There's an unpatched RCE in log4j1 that's existed since April 2017.  ](https://www.cvedetails.com/cve/CVE-2017-5645/)

kwwall:
@xeno6696 - I'm not really against this per se because @ricksouza also implemented it for JUL as well as for log4j 1.x. I consider the ;logging of the appname and username as a failed experiment. The thought was that multiple applications might all be logged to the same log file (e.g., multiple .war files in a .ear) so that the appname would be useful. In reality, that is rarely done and I doubt most even set it in the ESAPI.properties file. As for the current authenticated user name, the thought was that developers who used ESAPI would use ESAPI's Authenticator interface, but customize their own enterprise implementation. (The reference implementation is but a toy implementation that no one at all uses other than us for some unit testing.)

There are 2 reasons that I can think of rejecting this PR, but if it gets revised to address these things, then I would vote to accept it. My first reason to reject it is that there are no new unit tests to test it. Without unit tests, future changes could accidentally break it, so no JUnit tests then no merging of the PR.  My 2nd requirement if @ricksouza wants this merged is that it needs to be made to work with @jeremiahjstacey SLF4J support that he added. The reason for this, if we set up an expectation of something working for log4j and JUL as the logger implementation, then that is going to set up an expectation that the SLF4J logger implementation supports the identical functionality.

In the long term (ESAPI 3.x), I think we should _only_ support SLF4J and that this should be supported via SLF4J's MDC so that this could be supported via "pattern layouts". (I suspect that they didn't do that originally for log4j 1.x because JUL doesn't support MDC out-of-the-box; I'm pretty sure with JUL you'd have to create filters and/or formatters that are aware that you are using MDC. In short, it probably would be messy. But for ESAPI 3.x, we are not necessarily going to promise backward compatibility with ESAPI 2.x (not that we are going to try to intentionally break it either, but that's a discussion for elsewhere--like the mailing list).

But if this PR adds some JUnit tests that ensures that future changes don't break this functionality and makes it work with SLF4J, I think I would be willing to accept it as long we it doesn't add any new _runtime_ dependencies.

cristiantm:
Thanks for the feedbacks. On our case we are using Java Log implementation, not log4j 1.x, just implemented the support for both since we were working on it. In fact we did not even noticed if would be only log4j 1.x compatible (is it?! :)). 

I understand that you are unconfortable in that case. Maybe we should just keep JUL implementation and revert Log4j one? I also think it is interesting to add some note on ESAPI.properties, since the default one brings log4j as default logger implementation. It may look like an higher incentive to use it. 

Ragardind unit tests to cover this modifications, we will work on it. 

Ragarding a SLF4J implemenatation, we understand it currently does not have the same issue - it does not print the user and app info. In fact it does not use also the pre-existing ApplicationName setting. Could you explain in more details what you mean with "make it work with SLF4J" on that case? It seems that adding the app info and user info (defaulting to true) would break backwards compatibility on SL4J, while adding both defaulting to false (so it does not break SL4J) would break backward expectation on JUL/Log4j.

xeno6696:
> I understand that you are unconfortable in that case. Maybe we should just keep JUL implementation and revert Log4j one?

@cristiantm No, if Kevin is okay with leaving log4j changes in place, I can consent to that.  

> I also think it is interesting to add some note on ESAPI.properties, since the default one brings log4j as default logger implementation. It may look like an higher incentive to use it. This discussion makes me wonder that this is a relevant warning that could be added on the ESAPI.properties ("We recommend SLF4J because of..." and so on)

That's a fantastic suggestion, I should have thought of that years ago lol!

kwwall:
@cristiantm wrote:
>  understand that you are unconfortable in that case. Maybe we should just keep JUL implementation and revert Log4j one?

and

> Ragarding a SLF4J implemenatation, we understand it currently does not have the same issue - it does not print the user and app info. In fact it does not use also the pre-existing ApplicationName setting. Could you explain in more details what you mean with "make it work with SLF4J" on that case? It seems that adding the app info and user info (defaulting to true) would break backwards compatibility on SL4J, while adding both defaulting to false (so it does not break SL4J) would break backward expectation on JUL/Log4j.

No, my preference is that the JUL, the log4j 1.x, and the SLF4J implementations all work identically. I was not aware that the log format for SLF4J implementation was different than the other 2, but I really think that it ought to be. In fact, I think perhaps we ought to create a GitHub issue to correct that behavior if it is feasible. Conceptually it doesn't seem like it should be that hard, but I've never played around with it because we don't have a good full working example / tutorial on it (something that several on Stack Overflow have asked for.)

kwwall:
> > I also think it is interesting to add some note on ESAPI.properties, since the default one brings log4j as default logger implementation. It may look like an higher incentive to use it. This discussion makes me wonder that this is a relevant warning that could be added on the ESAPI.properties ("We recommend SLF4J because of..." and so on)
> 
> That's a fantastic suggestion, I should have thought of that years ago lol!

Yes, that would be a good idea I think, but I think to move people to the SLF4J implementation we need to support the same functionality as we have for the JUL and log4j 1.x implementations. If we did that, we could maybe even announce that we are going to deprecate the log4j 1.x implementation and replace it with the SLF4J implementation (using logback? or log4j 2.x??). But I think we ought to have the ability to make those SLF4J logs look like the default logs for JUL and log4j 1.x, which we apparently don't. IMO, of utmost importance is the ability to log the client IP address. There's a lot that a SIEM can do with that. Of secondary importance I think is the ability to get the user name logged. (I'm assuming that the SLF4J implementation does log the ESAPI event type since IIRC, that it is part of the Logger interface.and it would seem really weird not to log it. If not, I'd say that was second most important.)

ricksouza:
Hi to all,

I have started to implement the tests, but i would like to know your opinion. I saw that the JavaLogger and Logger4J tests are concentrated in printing logs using for example:

"testLogger.info"
"testLogger.debug"

But i did not find any test to verify for example if the Ip option is enable or not or if it is logging correctly. I was trying to find some example of how to test the app and user info to maintain the same pattern of tests. What do you suggest?

I was thinking to create a new JavaLogger and Log4JLogger test files, using a modified ESAPI.Properties (used on the tests) that have a default true value for app and user info. After that, i was thinking to use the setup to modify that and put both false and try to log infos, warning, errors and debugs.

cristiantm:
Hi all. Any thoughts on what test would we want to be added to this, sin it basically affects log formatting that is not tested at all currently?

kwwall:
You may want to wait until we merge @jeremiahjstacey PR that has some major
changes to the logging JUint tests. I've been too busy with Christmas
coming up to get it merged, but will try to get to it next week.

jeremiahjstacey:
@cristiantm @kwwall   I would like to just consume these changes in #532  and not worry about a merge.  Would that be Ok?
I've already built in placeholders for these components in all of the factories.  I would just need to update 3 classes, and take the configurations for the ESAPI.properites (though I think there are several more versions to update), and the DefaultSecurityConfiguration.

I believe I can have this integrated in short order and have all of the logging stuff done in one merge.

cristiantm:
Great! Thanks @jeremiahjstacey

kwwall:
@jeremiahjstacey Works for me as well if that's how you want to handle it.

-kevin
--

jeremiahjstacey:
Please review the last set of commits to #533 
These should account for the desired output from this review.   I would suggest closing this PR if the content is inline with the desired end state.

cristiantm:
Hi @jeremiahjstacey 

We reviewed the implementation and I understand there is a difference. 

Our implementation would have the following results:

LogUserInfo=true and LogAppInfo=true: [EVENT SUCCESS Anonymus:null@unknown -> 
    /com.sample.package.ABC][msg]
LogUserInfo=true and LogAppInfo=false: [EVENT SUCCESS Anonymus:null@unknown][msg]
LogUserInfo=false and LogAppInfo=true: [EVENT SUCCESS /com.sample.package.ABC][msg]
LogUserInfo=false and LogAppInfo=false: [EVENT SUCCESS][msg]

Your implementation "APP_INFO", that translates to "clientInfo" variable, is the portion after "@". Our implementation APP_INFO was the part after ->. We would like to hide also this part since it is already added on our appender/formatter. We included also some logic to only add "->" when its relevant. 

I think we should consider clarifying this expectations and meaning of this variables/configuration itens.

ricksouza:
> Please review the last set of commits to #533
> 
> * [f4921e3](https://github.com/ESAPI/esapi-java-legacy/commit/f4921e35d05056838ba173c2e442a521049795a8)
> * [617f373](https://github.com/ESAPI/esapi-java-legacy/commit/617f3735e3a79a8d1ae343e9a920e0eb905b642a)
> * [47b8929](https://github.com/ESAPI/esapi-java-legacy/commit/47b8929e518c9c45a04cde33fb967a0f06c13923)
> 
> These should account for the desired output from this review. I would suggest closing this PR if the content is inline with the desired end state.

Hi Jeremiah, actually we saw some differences between our implementation and what you did. Our objective was to remove some of the logs information that it was not important.

For example, our implementation was considering the following pattern:

"[" + typeInfo + getUserInfo() + " -> " + appInfo + "]

Therefore, the configuration could inform if it was necessary to log the UserInfo and AppInfo.

In your implementation, the pattern becomes:

[EVENT_TYPE USER_INFO:CLIENT_INFO -> SERVER_INFO]

You have converted the app_info to client_info, but we don't know exactly where the server_info comes from. Actually we thought it was not necessary to have more details logged. Our objective was to eliminate the "USER_INFO:CLIENT_INFO -> SERVER_INFO" in your implementation and the getUserInfo() + " -> " + appInfo in the old implementation.

Thus, it seems different from what we would like to log.

jeremiahjstacey:
I believe there is some confusion in the conversion.  

The implementation is not logging any additional information.  It has broken apart the pieces that compose what may be logged, but no new content has been added.

`"[" + typeInfo + getUserInfo() + " -> " + appInfo + "]`
- `typeInfo` is controlled by `EventTypeLogSupplier` (I refer to this as EVENT_TYPE)
- `getUserInfo()` creates a string that is `username:hostname@sessionId`.  I've deconstructed into two parts: 
  - `username` -> `UserInfoSupplier` (I refer to this as USER_INFO)
    - Property: `DefaultSecurityConfiguration.LOG_USER_INFO`
  - `host & sessionId` -> `ClientInfoSupplier` (I refer to this as CLIENT_INFO)
    - Property: `DefaultSecurityConfiguration.LOG_APP_INFO`
      - I'm going to rename this to `DefaultSecurityConfiguration.LOG_CLIENT_INFO`
      - I would also like to cascade this to the actual property value to be consistent.
- appInfo is being provided by the `ServerInfoSupplier` (I refer to this as SERVER_INFO)
  - Property: `DefaultSecurityConfiguration.LOG_SERVER_IP`
  - Property: `DefaultSecurityConfiguration.LOG_APPLICATION_NAME`

Please take a look at each of the Suppliers (and tests) to see what each section is logging.

The class `LogPrefixAppender` is responsible for concatenating all of the content together in the desired format.

Each of the '*LogFactory` implementations read the ESAPI configuration properties and create `LogPrefixAppender` implementations with the property values.


There is an inconsistency in the `LogPrefixAppender`.  It can be made smarter in what it logs when properties are set to false.  Reviewing the implementation, if all of the pertinent properties are set to false then it will still prefix `[TYPE_INFO :->]` to all logs.  I intend to follow up with this and improve this behavior.  I believe this correlates to what @cristiantm had identified.

The three things I intend to do are:
1. rename  `DefaultSecurityConfiguration.LOG_APP_INFO` to  `DefaultSecurityConfiguration.LOG_CLIENT_INFO`
2. rename the ESAPI.properties entry `Logger.AppInfo` to `Logger.ClientInfo`
3. Improve the logging output of `LogPrefixAppender` to remove the concatenation components when certain parts of the message are removed as outlined below:
  - `DefaultSecurityConfiguration.LOG_USER_INFO` = true
  - `DefaultSecurityConfiguration.LOG_CLIENT_INFO` = true
  - `DefaultSecurityConfiguration.LOG_SERVER_IP` = true
  - `DefaultSecurityConfiguration.LOG_APPLICATION_NAME` = true
    - [TYPE_INFO USER_INFO:CLIENT_INFO -> SERVER_INFO]
  - `DefaultSecurityConfiguration.LOG_USER_INFO` = true
  - `DefaultSecurityConfiguration.LOG_CLIENT_INFO` = false
  - `DefaultSecurityConfiguration.LOG_SERVER_IP` = true
  - `DefaultSecurityConfiguration.LOG_APPLICATION_NAME` = true  
    - [TYPE_INFO USER_INFO -> SERVER_INFO]
 - `DefaultSecurityConfiguration.LOG_USER_INFO` = false
  - `DefaultSecurityConfiguration.LOG_CLIENT_INFO` = true
  - `DefaultSecurityConfiguration.LOG_SERVER_IP` = true
  - `DefaultSecurityConfiguration.LOG_APPLICATION_NAME` = true
    - [TYPE_INFO CLIENT_INFO -> SERVER_INFO]
  - `DefaultSecurityConfiguration.LOG_USER_INFO` = false
  - `DefaultSecurityConfiguration.LOG_CLIENT_INFO` = false
  - `DefaultSecurityConfiguration.LOG_SERVER_IP` = true
  - `DefaultSecurityConfiguration.LOG_APPLICATION_NAME` = true
    - [TYPE_INFO  -> SERVER_INFO]
 - `DefaultSecurityConfiguration.LOG_USER_INFO` = false
  - `DefaultSecurityConfiguration.LOG_CLIENT_INFO` = false
  - `DefaultSecurityConfiguration.LOG_SERVER_IP` = false
  - `DefaultSecurityConfiguration.LOG_APPLICATION_NAME` = false
    - [TYPE_INFO]

cristiantm:
Ok, I think that, besides the confusion on communication, you understood everything.

With the 3 changes you cited it looks exactly as we expected/intended on this PR.

jeremiahjstacey:
Changes have been pushed up.  I found one additional correction to the EventTypeSupplier that defaults a null EventType to Logger.EVENT_UNSPECIFIED.   This should only impact the Log4JLoggerFactory, but the output would have resulted in empty brackets `[]` in that scenario.

cristiantm:
Looks great. Thanks!

jeremiahjstacey:
#533 has been merged which includes the required changes for this capability in the updated logging structures.  Capabilities have been verified by @ricksouza  and @cristiantm as fulfilling intended use cases.

Issue #521, opened 2019-10-07
JUnit test ValidatorTest.testIsValidSafeHTML() now failing


kwwall:
Reproduce by using JDK 8 and run the following commands:
```
$ rm -fr ./esapi-java-legacy
$ git clone https://github.com/ESAPI/esapi-java-legacy.git
$ cd ./esapi-java-legacy
$ mvn compile
$ mvn -Dtest=org.owasp.esapi.reference.ValidatorTest#testIsValidSafeHTML test
```
which will result in a failure at line 890 in ValidatorTest.java, which is this line:
`        assertTrue(instance.isValidSafeHTML("test", "Test. <script>alert(document.cookie)</script>", 100, false));`
The details of the failure, including what commit broke this test may be found [here](https://github.com/ESAPI/esapi-java-legacy/pull/514#issuecomment-527288355).

The "fix" for this **MUST NOT** _simply_ consist of changing this test. As part of commit [d869303](https://github.com/ESAPI/esapi-java-legacy/commit d869303/d86930362d703c790a18453882db1541c23f5c16), we have broken previously existing behavior, and even if that behavior was incorrect (and that could be argued since the behavior did not previously follow the documented Javadoc).

So we have two options: 1) roll back to the previous (and presumably incorrect?) behavior and "fix" by changing the Javadoc to agree with that previous behavior,  2) keep the current new (correct?) behavior as the default, but add some property to the ESAPI.properties that will allow ESAPI users to revert back to the prior behavior, 3) leave everything as-is and just announce this in the next release notes. I personally am adamantly against the third approach because few, if any read the release notes.

Before any "fixes" I would like a debate on what the correct / best approach should be.

Note: @davewichers apparently has not seen this under CentOS 7 or 8 (or whatever version he is running). I'm hoping he can confirm that.

-kevin

davewichers:
I'll be back at work on Weds and I can retest then to verify behavior for
this test on CentOS 8.

jeremiahjstacey:
I'm seeing that there are 2 test in the ValidatorTest related to changes in the HtmlValidationRule that are now failing.

```
testIsValidSafeHTML
testGetValidSafeHTML
```
Between these tests there are 5 Strings that now cause validation failures.  They are repeated multiple times in the two tests.

```
"Test. <script>alert(document.cookie)</script>"
"Test. <div style={xss:expression(xss)}>"
"Test. <s%00cript>alert(document.cookie)</script>"
[...]

Issue #509, opened 2019-08-02
HTMLValidationRule.getValid(String,String) does not follow documented specifications


ZiluckMichael:
The documentation for ValidationRule.getValid(String,String) states the following:
> Parse the input, throw exceptions if validation fails

However, `HTMLValidationRule.getValid(String,String)` instead calls `HTMLValidationRule.invokeAntiSamy`. If that method encounters any errors as part of the scan, it will simply log the errors and return the cleaned version of the HTML. What is desired instead, is an exception being thrown at that point that prints the error messages.

Links for ease of inspection:

- [ValidationRule.getValid(String,String) documentation](https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/ValidationRule.java#L10)
- [HTMLValidationRule.getValid(String,String) call](https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/reference/validation/HTMLValidationRule.java#L83)
- [HTMLValidationRule.invokeAntiSamy(String,String) ignoring the errors](https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/reference/validation/HTMLValidationRule.java#L117)

ZiluckMichael:
This one bug causes both #279 and #204

Issue #408, opened 2017-07-27
Update Travis CI configuration


augustd:
Travis.yml (https://github.com/ESAPI/esapi-java-legacy/blob/develop/.travis.yml) currently points to bkimminich/esapi-java-legacy but that is not a valid repository on https://travis-ci.org: 

https://travis-ci.org/bkimminich/esapi-java-legacy
"The repository at bkimminich/esapi-java-legacy was not found."

kwwall:
So does anyone want to "host" a Travis CI instance? Bjoern (@bkimminich) was kind enough to offer to set up the site originally. We probably should have either that go back to the Cloudbees Jenkins CI that preceded it. (The later should be referenced in an older version of README.md if anyone cares to go looking.)

bkimminich:
The ESAPI org user must register on Travis-CI, set up a build and then simply replace my username with "ESAPI" in the configuration. If there are problems, I'm happy to help.

augustd:
I have a Travis account, but it only allows me to access my OSS projects under https://github.com/augustd

Presumably, whoever owns the @ESAPI user account will have to setup Travis.

kwwall:
I guess that would be me, but looks like Travis is insisting on using my GitHub account to authenticate, which is maybe acceptable (I'd prefer a separate, non-associated account), but it seems like Travis is requesting excessive GitHub privs to me (*manage* deployments, *write* commit status, *write* access to webhooks and services, That's _way_ more than I'm comfortable with. Does it have to be the OWNER of the @ESAPI account? If not, I'll create a more restricted account and maybe use it to execute CI builds on Travis. Surely there are more free Jenkins or Hudson services that would allow more or less the equivalent without such excessive permissions, isn't there? (Cloudbees seems to be a for-pay service, although they may have an exception for FOSS; I've not looked that closely.)

kwwall:
I'm looking for someone who would like to take ownership of this. I really don't care which Continuous Integration tool is used (Jenkins, Hudson, Travis, etc.) but as you can see from my comments, I have objections giving Travis full admin GitHub rights. So maybe someone who knows more about CI could make a recommendation or what ESAPI should use and how we should set it up.

kwwall:
@jackycct Could you look into this? I don't really care that much WHAT free hosted CI service that we use but we ought to be using one that at least compiles the code and runs the JUnit tests.
[...]

PR #535, opened 2020-01-21
Issue 521

kwwall:
This is to address issue #521 
Note that the "fix" is rather a bit of a kludge. A new property name was introduced, but I'm not very comfortable with the actual property name. (My patience along with my cleverness ran out after about five minutes of failing to come up with anything better.) So please at least review that part and see if you can come up with anything better.

Unrelated to issue 521 is a semi-formal proclamation of the ESAPI deprecation policy that I added to ESAPI's README.md file. I probably should have pulled it into a separate PR, but since it's just a text file, I figured, meh, why bother?


src/main/java/org/owasp/esapi/reference/validation/HTMLValidationRule.java
```diff
@@ -97,8 +97,33 @@ public String sanitize( String context, String input ) {
 		return safe;
 	}
 
+    /**
+     * This admittedly is an UGLY hack to ensure that issue 509 and it's corresponding
+     * fix in PR #510 does not break existing developer's existing code. Full
+     * details are described in GitHub issue 521.
+     */
+    private boolean ignoreIssue509Fix() {
+        boolean ignore = true;
```
jeremiahjstacey: I don't feel strongly, but I feel like this should default to the "correct" behavior and default to false?  Otherwise we risk new users becoming accustomed to the incorrect behavior?

I suggest making this property "Opt in to be unsafe"

jeremiahjstacey: I think this should be false by default and throw the exception in the implementation?

configuration/esapi/ESAPI.properties
```diff
@@ -499,3 +499,13 @@ Validator.DirectoryName=^[a-zA-Z0-9:/\\\\!@#$%^&{}\\[\\]()_+\\-=,.~'` ]{1,255}$
...
+
+# This is a workaround to make sure that a commit to address GitHub issue #509
+# doesn't accidentally break someone's production code. So essentially what we
+# are doing is to default back to the previous buggy (but, likely by now, expected)
+# behavior. More defaults about this are written up under GitHub issue #521 and
+# the pull request it references. Future major releases of ESAPI (e.g., ESAPI 3.x)
+# will not support this previous buggy behavior, but it will remain for ESAPI 2.x.
+# Set this to 'false' if you want the originally intended behavior of throwing
+# that was fixed via issue #509.
+Validator.ValidationRule.getValid.ignore509Fix=true
```
jeremiahjstacey: I think the documentation and the implementation are not aligned.  If this is "true" then the implementation logs and returns the empty string.  If  the value is false, then the implementation adheres to the API and docs and throws the expected Exception.

xeno6696: Yeah, the semantics here are a little weird.  And I've in multiple instances, run into cases where variable names regarding booleans can lead us down faulty paths.

src/main/java/org/owasp/esapi/reference/validation/HTMLValidationRule.java
```diff
@@ -97,8 +97,33 @@ public String sanitize( String context, String input ) {
...
+                           org.owasp.esapi.reference.DefaultSecurityConfiguration.VALIDATOR_IGNORE509 +
+                           " either unset or not set to a boolean value. Assuming 'true'.", cex);
+        }
+
+        return ignore;
+    }
+
 	private String invokeAntiSamy( String context, String input ) throws ValidationException {
-		// CHECKME should this allow empty Strings? "   " us IsBlank instead?
+		// CHECKME should this allow empty Strings? "   " use IsBlank instead?
```
xeno6696: Agreed.  IsBlank.

src/main/java/org/owasp/esapi/reference/validation/HTMLValidationRule.java
kwwall: My rationale for making this false if because I wanted to ensure backwards compatibility (i.e., not breaking someone's production code) if the updated their ESAPI jar to 2.2.1.0 (or whatever we name the next release) but fail to update their production ESAPI.properties file, which I think is probably a pretty common thing to do. E.g., they may have ESAPI.MasterKey / ESAPI.MasterSalt set or have done other extensive customization to it so that a wholesale copying of the updated ESAPI.properties file (which you can't download from Maven Central so is a bit harder to get anyway) so they figure it is safer to leave as-is, trusting that won't break anything. And generally they would be right. So it they take that approach, at best they may look for new properties to add in, but they very well could miss one, especially given it's dubious name of "Validator.ValidationRule.getValid.ignore509Fix". Also, I'm still not 100% convinced that throwing an exception is the correct behavior. We have no way of knowing if the Javadoc just wasn't wrong and since AntiSamy still ends up cleanses these cases, not throwing an exception here isn't really introducing any security risk that I can see here.

configuration/esapi/ESAPI.properties
kwwall: I partly think that the semantics are weird here because, let's face it, this property name sucks dirt. Essentially what the property should be (but in my opinion is even worse) would be something like:
`Validator.ValidationRule.getValid.have_esapi_act_like_it_used_to`
or
`Validator.ValidationRule.getValid.ensure_backward_compatibility_even_though_that_probably_was_a_bug`
which which case 'true' would make more semantic sense.  I think the 'ignore' part of the property name along with sets up a mental dichotomy because it sort of implies a negation that leads to confusion. I think the only way to clear it up may be to come up with a better property name.

So what is a more suitable property name that will cause less confusion? I'm open to suggestions.

@jeremiahjstacey wrote:

> The value of false allows the implementation to act like the other validation rules, and throw an exception if errors are found while validating incoming HTML content.

I think there is one major difference between this and the other rules for validating HTML content and that is those other rules generally don't try to "cleanse" or "sanitize" the output, whereas AntiSamy does. And that's another thing that gives me a little bit of uncertainty about whether the original Javadoc was correct or not.

src/main/java/org/owasp/esapi/reference/validation/HTMLValidationRule.java
kwwall: This was just a fixing a typo in the spelling. I'm not against changing this necessarily, but at this point, that could go beyond the original intent of this issue of just restoring the behavior to what it was before GitHub issue #509 was addressed in the PR that broke our JUnit tests.

If we want this changed, I think that someone ought to create a separate GitHub issue and do it under a separate PR. That way if we need to undo it for some reason (like, it turns out that it could break code again--and I doubt we have a test case for this), then it will make it easier to undo. Does that work for the two of you?

Also, BTW, there are CHECKME, DISCUSS, OPEN ITEM, and TODO comments littered throughout ESAPI that I have wanted someone to review and provide feedback on, but I've never previously had any takers. I probably added the majority of them. If either of you are feeling ambitious, don't stop at this one. :)

jeremiahjstacey:
I did not understand fully where we would end up with this issue when we started discussing it last year. I believe I can see the result we're looking for with the effort.  I think I see what we're looking for now, and rather than try to talk through it I just created a branch and stubbed the functional behavior:
https://github.com/jeremiahjstacey/esapi-java-legacy-dev/commit/b498ee46b72714b64af7ff15c7cb0da20080eda2

Forgive the lack of testing, but my intent is to allow HTMLValidationRule to comply with the API while allowing the implementation of the DefaultValidator to alter the response to the exception.   

This implementation also changes the boolean property into a named behavioral strategy, which I think adds more clarity.

configuration/esapi/ESAPI.properties
kwwall: @jeremiahjstacey, @xeno6696  - Still looking for better suggestion for this property name. I can easily rewrite the code, but this is what is most visible to ESAPI users and we need to get this right before we do a release. And I'd like to get some urgency around this PR because this will fix the broken tests in our test suite which we need if we are going to get issue #408 closed out that @kerberosmansour is working on.

I really don't like this property name--at all--but I didn't want it to hold me up from finishing the rest. If we can come up with a name that doesn't imply a double negative, I will gladly adjust the code to these suggestions that you had.

jeremiahjstacey:
@kwwall  If you have not already, please take a look at the strategy-pattern implementation I linked in https://github.com/jeremiahjstacey/esapi-java-legacy-dev/commit/b498ee46b72714b64af7ff15c7cb0da20080eda2

I believe this resolves the property name issue by allowing the user to set the "HTML Validation Mode" in ESAPI.properties.

```
#Clean HTML returns a blank string on ValidationException. Consistent Legacy Support
Validator.htmlValidationMode=cleanhtml
#Safe HTML allows Validation Exceptions to propagate through DefaultValidator HTML checks.
#Validator.htmlValidationMode=safehtml
```

kwwall:
Replying to @jeremiahjstacey comments in https://github.com/ESAPI/esapi-java-legacy/pull/535#issuecomment-581073764 ...

I renamed the property to 'Validator.HtmlValidationAction', which must be set to 'clean' (which provides the legacy behavior of AntiSamy simply sanitizing the unsafe HTML input, logging the problem, and then returning the sanitized output (which is how it used to work; it didn't just return an empty string it returned the sanitized string), or if the property is set to 'throw' it will throw a ValidationException. I set the default in ESAPI.properties to 'throw', but configured it that if the property name is not encountered or if it is found but set to neither 'clean' or 'throw', the legacy behavior is taken (i.e., as if it was set to 'clean').

I did not use your "strategy pattern" implementation for several reasons, but the main one being is that the "forces" behind the GoF strategy pattern was when you wished to support many different strategies, possibly some of them even unforeseen at the moment. Here, we instead only have to possible choices...the new behavior that was addressed as a fix to issue #509 and the legacy behavior which we pretty much are only keeping around so not to break any client code using ESAPI.

So a re-review is probably in order if you would please. The one test still needs cleaned up and I should probably rename HTMLValidationRuleLogsTest.java to HTMLValidationRuleCleanTest.java and one of the new tests (forget which one off the top of my head) should be switched to JUnit 4. I just ran out of energy that night...

jeremiahjstacey:
@kwwall I agree with your assessment and appreciate the consideration in the implementation updates.  I have no additional concerns.  Once you're comfortable with the test updates you'd like to make I will merge these changes.  (Pending agreement from @xeno6696 )

kwwall:
@jeremiahjstacey @xeno6696 - JUnit test updates made and pushed.

PR #539, opened 2020-02-09
upgrade for convergence

wiiitek:
I have seen the commented text next to `powermock-api-mockito2` dependency in [pom.xml](https://github.com/ESAPI/esapi-java-legacy/blob/develop/pom.xml#L322) and I assume that we would like to have 100% convergence.

This pull request adds a rule for `maven-enforcer-plugin` that checks that.

I needed to upgrade `powermock-reflect` from 2.0.0 to 2.0.2 (didn't check if that works with JAVA 7).

I also merged two executions of the plugin into one:

1. if there is only one execution then the *id* is not needed
1. `<fail>` parameter for this plugin has **true** as default value

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,7 +364,7 @@
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-reflect</artifactId>
-            <version>2.0.0</version>
+            <version>2.0.2</version>
             <scope>test</scope>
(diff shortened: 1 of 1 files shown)


xeno6696:
@kwwall No issues here.  Will take the executive decision to just merge this one but feel free to stop me if you think its necessary!

Issue #542, opened 2020-05-12
Write up ESAPI release notes for planned 2.2.1.0 release


kwwall:
Pattern after the 2.2.0.0 release notes in:
https://github.com/ESAPI/esapi-java-legacy/blob/develop/documentation/esapi4java-core-2.2.0.0-release-notes.txt

PR #543, opened 2020-05-12
Release notes for 2.2.1.0

sempf:
I have no idea how to associate this with Issue #542  but if I figure it out I will do it.


kwwall:
@sempf - That's okay. I probably should leave issue #542 open anyway until I fill in all the TBD sections. But thanks for doing this.

PR #550, opened 2020-06-23
Fixed #522 by updating Javadoc.  

### LINKED ISSUE #522: javadoc corrections for Encoder.canonicalize()
Hi! I'm trying to understand how exactly the method `Encoder.canonicalize()`  works and for that I'm reading the javadocs from here:

- https://static.javadoc.io/org.owasp.esapi/esapi/2.2.0.0/org/owasp/esapi/Encoder.html#canonicalize-java.lang.String-boolean-boolean-

But when testing some code examples, I found things that I don't really understand; for example:

> In ESAPI, the Validator uses the canonicalize method before it does validation. So all you need to do is to validate as normal and you'll be protected against a host of encoded attacks.
>      String input = request.getParameter( "name" );
>      String name = ESAPI.validator().isValidInput( "test", input, "FirstName", 20, false);

If I read correctly, `ESAPI.validator().isValidInput()` returns a `boolean`, not a `String`. I'm guessing that the correct method to call in the last line of the example should be...`ESAPI.validator().getValidInput()` ?

And the same would apply here:

> However, the default canonicalize() method only decodes HTMLEntity, percent (URL) encoding, and JavaScript encoding. If you'd like to use a custom canonicalizer with your validator, that's pretty easy too.
>      ... setup custom encoder as above
>      Validator validator = new DefaultValidator( encoder );
>      String input = request.getParameter( "name" );
>      String name = validator.isValidInput( "test", input, "name", 20, false);

Also, in this example:

> This method allows you to override the default behavior by directly specifying whether to restrict multiple or mixed encoding. Even if you disable restrictions, you'll still get warning messages in the log about each multiple encoding and mixed encoding received.
>      // disabling strict mode to allow mixed encoding
>      String url = ESAPI.encoder().canonicalize( request.getParameter("url"), false, false);

Shouldn't the correct call be `ESAPI.encoder().canonicalize( request.getParameter("url"), true, true);`

Am I getting something wrong?

Thanks!

xeno6696:


xeno6696:
I seem to have stepped in it now.  I might have to back all that out for some changes I had never pushed and merged to my own fork.  =-(  

@kwwall might be shorter to just have you copy my javadoc comments I added here and I'll just start with a clean repo since I'm not even sure what I was trying to change 10months ago.

kwwall:
Yeah, let's just do that. Either point me to your commit or package your
changes and email them to me.

-kevin
--

xeno6696:
@kwwall here's the patch.  I'll take this moment to close the PR and then fix my crap.   
[commit-c460755.zip](https://github.com/ESAPI/esapi-java-legacy/files/4822870/commit-c460755.zip)

kwwall:
@xeno6696 - Thanks. That was pretty painless. Added it to the PR I'm preparing.

PR #551, opened 2020-07-02
Misc cleanup

kwwall:
1) Update to pom.xml to 2.2.1.0-RC1 and dependency version updates to address known CVEs.
2) Extensive Javadoc changes, especially in ESAPI Encoder.
3) Misc cleanup from issues identified by LGTM. (Many more remain.)
4) Update README.md with new OWASP wiki link and trademark info.


src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java
```diff
@@ -681,15 +681,15 @@ public String getServerName() {
...
-	public int getServerPort() {
-		int port = getHttpServletRequest().getServerPort();
-		if ( port < 0 || port > 0xFFFF ) {
-			logger.warning( Logger.SECURITY_FAILURE, "HTTP server port out of range: " + port );
-			port = 0;
-		}
-		return port;
-	}
- 	
+    public int getServerPort() {
```
xeno6696, line 684: I hate to ask this but with the tab to space transformation in this file alone, might it be worthwhile once we agree we're at the "end" of this release that we apply it to the whole project and then create a single commit with one change that is ONLY whitespace?

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -107,16 +109,42 @@
...
+ *      &lt;/script&gt;alert(1)&lt;/script&gt;
+ * </pre>
+ * or similiar simplistic XSS attack payloads and if that is properly encoded
+ * (or, you don't see an alert box popped in your browser), you consider it
+ * "problem fixed", consider the unit testing sufficient. Unfortunately, that
+ * minimalist testing may not always detect places were you used the wrong decoder. You need to do better.
+ * Fortunately, the aforementioned link,
+ * <a href="https://arxiv.org/pdf/1804.01862.pdf" target="_blank"
+ * rel="noopener noreferrer">
+ * Automated Detecting and Repair of Cross-SiteScripting Vulnerabilities through Unit Testing</a>
```
xeno6696, line 143: @jeremiahjstacey should we resurrect our old idea of a security-based fuzzing JUnit framework?  This part of the comments reminds me of some of our old talks about that.  (For @kwwall 's sake it amounts to taking a rather large collection of XSS attacks and automating attacking parameters with them.)

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -107,16 +109,42 @@
...
+ * provides some insight. You may also wish to look at the
+ * <a href="https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/test/java/org/owasp/esapi/reference/EncoderTest.java"
+ * target="_blank" rel="noopener noreferrer">ESAPI Encoder JUnittest cases</a>.
+ * If you are really ambitious, an excellent resource for XSS attack patters is
```
xeno6696: Should we also discuss SecLists and FuzzDB?

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -298,54 +332,55 @@
...
+     * directly into script code itself, as no amount of encoding will prevent attacks there.
+     * 
+     * This method is not recommended as VBScript is only supported by Internet Explorer
+     * 
+     * @param untrustedData 
+     *      the untrusted data to output encode for VBScript
+     * 
+     * @return the untrusted data safely output encoded for use in a use in VBScript
+     */
+    String encodeForVBScript(String untrustedData);
```
xeno6696, line 355: Nice switch.

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -365,171 +400,173 @@
...
-	 */
-	String encodeForXML(String input);
+    /**
+     * Encode data for use in an XML element. The implementation should follow the <a
+     * href="http://www.w3schools.com/xml/xml_encoding.asp">XML Encoding
+     * Standard</a> from the W3C.
+     * <p>
+     * The use of a real XML parser is strongly encouraged. However, in the
+     * hopefully rare case that you need to make sure that data is safe for
+     * inclusion in an XML document and cannot use a parse, this method provides
```
xeno6696, line 471: "parse" should be "parser"  ...?

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -365,171 +400,173 @@
...
+     * a safe mechanism to do so.
+     * 
+     * @see <a href="http://www.w3schools.com/xml/xml_encoding.asp">XML Encoding Standard</a>
```
xeno6696: Given my criticism of w3schools in the list email, ought we change this to the RFC standard or do we have some space where I could perhaps discuss this and related topics for reference here?  This week at work I got hit with a vendor trying to squeak by saying "*" was OK in a URI because the URI spec said so... when the real world is a bit more complicated.  There's a spec for the URI, and another spec for the HTTP protocol... I can see this mistake being repeated MANY times.

xeno6696:
So a couple of minor comments posted above, but I have no compunctions against having those hold up an RC release.  If I hear nothing else from @jeremiahjstacey before midday Saturday I'll merge this PR.

src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java
kwwall: You sure you referenced the correct file here? I just checked and I see no tabs at all in src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java.

kwwall: @xeno6696 - BTW I usually fix these by running ':retab' in gvim which converts all the tabs to the appropriate # of spaces with the tabsize set to 4 spaces.

src/main/java/org/owasp/esapi/Encoder.java
kwwall: I often had thought it would be useful to take many of our JUnit tests and make them data-driven so we could just put the test data and the expected outcomes into a file. If we did that, it would make fuzzing easier. I have wanted to do this for years for the Encryptor tests so I could take a more standard cryptographic test vectors approach. It makes it really easy to add new test cases. To my knowledge, the only place we are using an approach like that is with all those tests that you (Matt) added to testing the URL validation / canonicalization tests.

src/main/java/org/owasp/esapi/Encoder.java
kwwall: Most of this Encoder Javadoc originated on my running across https://arxiv.org/pdf/1810.01017.pdf, which was a usability study on the ESAPI Encoder. I have been interacting with the 2 authors on Google Drive commenting on their paper and trying to gain insight as to how to improve ESAPI Encoder to make it more noob friendly. They had remarked that students had no idea how to test to see if then had secured their code correctly, thus these specific additions. So that is the target audience with these javadoc comments on testing.

src/main/java/org/owasp/esapi/Encoder.java
kwwall: The discussion with the authors of the aforementioned arXiv paper where a commonly documented mistake was that participants were trying to use ESAPI Encoder's output encoding on the input (e.g., request parameters) rather than doing output encoding on the output side. I then noticed that most of the parameters were all called 'input' (which I switched to 'untrustedData' and then changed the @return for most of the HTTP related output encoders.

src/main/java/org/owasp/esapi/Encoder.java
kwwall: Good catch. Will upload a new version in a few minutes.

src/main/java/org/owasp/esapi/Encoder.java
kwwall: Yep; I changed some of them, but looks like I missed this one. Good eye.

kwwall: Ah, never mind. I did originally miss it, but then later corrected that one. (Note the 'Outdated' label next to the file name.)

src/main/java/org/owasp/esapi/Encoder.java
xeno6696: Yep we're tracking together.  In the past we built a custom JUnit test that started with an input file that collected *all* the attack inputs I could find, and then would include an expected value as to whether or not that attack should pass or fail whatever filter we were working on.  At the time, we had talked about creating a more generalized framework that folks could drop into their own codebase.  The issue here is that it requires a great deal of careful judgment and to be blunt, you can't trust THAT to just any programmer off the street.

xeno6696:
In line 131 and 132 is where I see an initial swap from tabs to spaces.  I wasn’t commenting here that I FOUND tabs, but it seemed suspect that it would only be in a handful of places… just wanted to make sure that we didn’t need to do something more global.  Sounds like you have this one under wraps however, so I’ll shut my grande yapper…   Sent from Mail for Windows 10 From: Kevin W. WallSent: Thursday, July 2, 2020 8:58 PMTo: ESAPI/esapi-java-legacyCc: Matt Seil; CommentSubject: Re: [ESAPI/esapi-java-legacy] Misc cleanup (#551) @kwwall commented on this pull request.In src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java:> @@ -681,15 +681,15 @@ public String getServerName() {      * HttpServletRequest after parsing and checking the range 0-65536.      * @return The local server port      */-       public int getServerPort() {-              int port = getHttpServletRequest().getServerPort();-              if ( port < 0 || port > 0xFFFF ) {-                      logger.warning( Logger.SECURITY_FAILURE, "HTTP server port out of range: " + port );-                      port = 0;-              }-              return port;-       }-       +    public int getServerPort() {You sure you referenced the correct file here? I just checked and I see no tabs at all in src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java.—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or unsubscribe.

kwwall:
Okay; having a hard time reading your comment with all the CSS and all, but based on what you described, I probably aready ran ":retab" from gvim on that file. That would do what you described and would explain it.

kwwall:
@davewichers One of your two commits to pom.xml (probably commit 807adf7) is not causing 'mvn site' to fail when it runs Dependency Check. See if you can reproduce it. If you can't reproduce the error, let me know and I will run it, save the output, and either attach it here or email it to you.

davewichers:
Works for me I’m pretty sure. Send me the error.

kwwall:
Here is the error that I got from dependency check the first time (using Java 8 and Maven 3.6.0 on Linux Mint 19.2).
Oddly I tried it again and it now seems to be working. May have had to do with something incompatible in the Dependency Check cache from an earlier Dependency Check version???
[mvn-site-error.txt](https://github.com/ESAPI/esapi-java-legacy/files/4873725/mvn-site-error.txt)

kwwall:
@xeno6696 - I cleaned up the merge conflict in case that was keeping you from doing the merge. Also, don't let the 'mvn site' problems hold this up. @davewichers is looking into it, but as I mentioned above, it now seems like a temporary glitch.

PR #553, opened 2020-07-03
Fix for GitHub Issue 552

### LINKED ISSUE #552: Rewrite implementation of some ESAPI classes to remove Java 8 dependencies
The current minimal baseline for ESAPI 2.2.0.0 is supposed to be Java 7. While the ESAPI pom.xml specifies <target> and <source> to use 1.7, we've been a bit lazy and haven't bothered to specify a '-bootclasspath' that refers to a Java 7 rt.jar file. As a result, we ended up unintentionally ended up using a few Java 8 (and later) packages. Those need to be removed if ESAPI can be used with Java 7.

kwwall:
Fix #552 
Note I did not really want to include documentation/esapi4java-core-2.2.1.0-release-notes.txt in this PR, but for some reason, git always thinks it has been modified (maybe because of end-or-line terminator?). Anyway, you can ignore that. Aside from a 2 line change it is the same as what I have in the 'misc-cleanup' PR.


jeremiahjstacey:
Looks good to me!  Thanks again!

Issue #560, opened 2020-07-22
Could not initialize class org.owasp.esapi.logging.java.JavaLogFactory (ESAPI 2.2.1.0)


elangoravi:
Hi Team,

We've been migrating from ESAPI 2.1.0.1 version to the latest version 2.2.1.0

When I checked the release notes, I could understand the logging implementations/structure have undergone some major changes.

All I did was include the latest version and changed the property as below 
**ESAPI.Logger=org.owasp.esapi.logging.java.JavaLogFactory**

And it throws the below exception with our existing code:
Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.owasp.esapi.logging.java.JavaLogFactory
        at java.lang.Class.forName0(Native Method)
        at java.lang.Class.forName(Class.java:264)
        at org.owasp.esapi.util.ObjFactory.loadClassByStringName(ObjFactory.java:158)
        at org.owasp.esapi.util.ObjFactory.make(ObjFactory.java:81)
        at org.owasp.esapi.ESAPI.logFactory(ESAPI.java:139)
        at org.owasp.esapi.ESAPI.getLogger(ESAPI.java:155)
        at org.owasp.esapi.reference.DefaultHTTPUtilities.<init>(DefaultHTTPUtilities.java:137)
        at org.owasp.esapi.reference.DefaultHTTPUtilities.getInstance(DefaultHTTPUtilities.java:93)

(checked the release notes of 2.2.0.0 and added the necessary properties and this version is working fine with our existing code)

Apologies if I'm being naive/blunt.

Many thanks

jeremiahjstacey:
There are 2 cases where I've seen NoClassDefFoundErrors previously.
The easy one would be if the path to the JavaLogFactory was off.  This is not the case, your definition is correct from what I'm seeing.  `org.owasp.esapi.logging.java.JavaLogFactory`
The harder one to debug is a failure during static initialization.  There are several static blocks in the class that attempt to initialize the instance.
https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java

They rely on several properties being defined in ESAPI.properties.  

The properties are defined in the DefaultSecurityConfigurationClass
https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
```
[...]

PR #562, opened 2020-07-24
Issue #560 JUL fixes

jeremiahjstacey:
Updating the property loading to account for the possible null stream.  Updated exception handling to rethrow ConfigurationException on null stream or IOException from load attempt.

Default content for esapi-java-logging.properties is added to the configuration directory


kwwall:
Everything looks good. Thanks @jeremiahjstacey for attending to this so quickly.

Issue #571, opened 2020-09-13
Logger.always() fails to log all the time when ESAPI is using org.owasp.esapi.logging.java.JavaLogFactory


kwwall:
Tested in ESAPI 2.2.1.1

The _intent_ of the two `Logger.always()` methods is to _always_ do logging regardless of what the log level is set to in your logging configuration. However, if ESAPI is using the (current) default JUL as logging via setting

`ESAPI.Logger=org.owasp.esapi.logging.java.JavaLogFactory`

in **ESAPI.properties** and in **esapi-java-logging.properties**, the log level is set to 'INFO', then this will not produce any output:

`        ESAPI.log().always(Logger.EVENT_SUCCESS,"This should ALWAYS appear!!!");`

However, changing the entry
`java.util.logging.ConsoleHandler.level = INFO`
to
`java.util.logging.ConsoleHandler.level = ALL`

will cause the Logger.always() output to correctly log.

kwwall:
@jeremiahjstacey - If you have some time, could you take a quick look at this? If you don't have time to fix it and submit a PR, then maybe just point me to the proper place to fix it and outline what needs to be done? Thanks.

jeremiahjstacey:
I'm debugging the method that you've outlined above and I think I have found the root of the issue.  I believe that the underlying Java Logger is filtering out java.util.Level.ALL by default. 

As the code exists now:
- The ESAPI Level.ALL maps to the java.util.Level.ALL
- ESAPI Logger.always does always propagate the log event to the delegate Logger.
- The Java Logger implementation defaults to INFO.  The implementation of java.util.Logger.isLoggable filters out Level.ALL when set to Level.INFO

In the source I have for Java, the java.util.Logging.isLoggable method is as follows:
```
  /**
     * Check if a message of the given level would actually be logged
     * by this logger.  This check is based on the Loggers effective level,
     * which may be inherited from its parent.
     *
     * @param   level   a message logging level
     * @return  true if the given message level is currently being logged.
     */
    public boolean isLoggable(Level level) {
        if (level.intValue() < levelValue || levelValue == offValue) {
            return false;
        }
        return true;
    }
```
Note the first part of the if block which compares the int values of the java.util.Logging Level against the current system configuration.   The int values are determined by the java.util.Level Enumeration:
```
public static final Level INFO = new Level("INFO", 800, defaultBundle);
public static final Level ALL = new Level("ALL", Integer.MIN_VALUE, defaultBundle);
```
[...]

PR #572, opened 2020-09-13
Issue 571: JavaLogFactory ALWAYS

jeremiahjstacey:
https://github.com/ESAPI/esapi-java-legacy/issues/571#issuecomment-691674605

Adding implementation noted in the comment listed above.

Caveat to this implementation is that if the Java Logging configuration is set to OFF then no logging will occur.  I don't think there's much that can be done about this particular use case.

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
+++ b/src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
@@ -31,6 +31,11 @@ public class ESAPICustomJavaLevel extends Level {
     public static final Level ERROR_LEVEL = new ESAPICustomJavaLevel( "ERROR", Level.SEVERE.intValue() - 1);
 
     /**
+     * Defines a custom level that should result in content always being recorded, unless the Java Logging configuration is set to OFF.
+     */
+    public static final Level ALWAYS_LEVEL = new ESAPICustomJavaLevel( "ERROR", Level.OFF.intValue() - 1);
+    
+    /**
      * Constructs an instance of a JavaLoggerLevel which essentially provides a mapping between the name of
      * the defined level and its numeric value.
      * 
--- a/src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
+++ b/src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
@@ -18,7 +18,7 @@ package org.owasp.esapi.logging.java;
 import java.util.logging.Level;
 
 /**
- *  A custom logging level defined between Level.SEVERE and Level.WARNING in logger.
+ *  Definitions of customized Java Logging Level options to map ESAPI behavior to the desired Java Log output behaviors.
  */
 public class ESAPICustomJavaLevel extends Level {
 
@@ -33,7 +33,7 @@ public class ESAPICustomJavaLevel extends Level {
     /**
      * Defines a custom level that should result in content always being recorded, unless the Java Logging configuration is set to OFF.
      */
-    public static final Level ALWAYS_LEVEL = new ESAPICustomJavaLevel( "ERROR", Level.OFF.intValue() - 1);
+    public static final Level ALWAYS_LEVEL = new ESAPICustomJavaLevel( "ALWAYS", Level.OFF.intValue() - 1);
     
     /**
      * Constructs an instance of a JavaLoggerLevel which essentially provides a mapping between the name of
--- a/src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
+++ b/src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
@@ -20,7 +20,7 @@ import java.util.logging.Level;
 /**
  *  A custom logging level defined between Level.SEVERE and Level.WARNING in logger.
  */
-public class ESAPIErrorJavaLevel extends Level {
+public class ESAPICustomJavaLevel extends Level {
 
     protected static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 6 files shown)


kwwall:
Maybe note that (log level for JUL set to OFF) that always() methods will
not log in the Javadoc for the 2 JavaLogger always() methods. (Or in the
Logger.always() Javadoc if it applies to the other loggers.)

-kevin
--

jeremiahjstacey:
From what I'm seeing, both Logback and Log4J2 have similar configuration options for ALL/OFF.

I will update the ESAPI Logger always() method Javadoc to state that if the underlying logging configuration is disabled that no logging will occur.

http://logback.qos.ch/apidocs/ch/qos/logback/classic/Level.html
https://howtodoinjava.com/log4j/logging-levels-in-log4j/
https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Level.html

kwwall:
Thanks. Sounds good.

-kevin
--

src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
```diff
@@ -20,15 +20,15 @@
 /**
  *  A custom logging level defined between Level.SEVERE and Level.WARNING in logger.
  */
-public class ESAPIErrorJavaLevel extends Level {
+public class ESAPICustomJavaLevel extends Level {
```
jeremiahjstacey, line 23: I should update the class documentation to reflect the updated content.

src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
```diff
@@ -30,6 +30,11 @@
      */
     public static final Level ERROR_LEVEL = new ESAPICustomJavaLevel( "ERROR", Level.SEVERE.intValue() - 1);
 
+    /**
+     * Defines a custom level that should result in content always being recorded, unless the Java Logging configuration is set to OFF.
+     */
+    public static final Level ALWAYS_LEVEL = new ESAPICustomJavaLevel( "ERROR", Level.OFF.intValue() - 1);
```
jeremiahjstacey, line 36: Naming Error.  Should be "ALWAYS" and not "ERROR".  Copy/Pasta mistake

src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
kwwall, line 23: I meant to comment on this before. It's the main reason why I haven't accepted this PR. While I agree with the intent and this name does better reflect what this class is about, I think it is best **NOT** to change this class name because it could _potentially_ break someone's code that uses ESAPI. While I think it is _highly_ unlikely, someone could have been directly using `ESAPErrorJavaLevel`. (After all, it is a _public_ class library that does not hide components via Java modules and there is no "Do not use" warning in the class' Javadoc.) So, if we were to remove this, I'd have to call special attention to it in the next release notes (assuming that I'd remember).

If you really want to get rid of this class, then mark `ESAPIErrorJavaLevel` as deprecated, note in the Javadoc that it's being replaced by `ESAPICustomJavaLevel` and then we will mark it for removal in either 2 years after the next release or in the next major release (presumably 3.0, whenever that is). Then you can either copy/paste `ESAPIErrorJavaLevel` to `ESAPICustomJavaLevel` and tweak as needed (which is awkward, but probably best) or use delegation to that class. (I'm thinking that if you deprecate `ESAPIErrorJavaLevel` and then extended it as in:

`public class ESAPICustomJavaLevel extends ESAPIErrorJavaLevel`

the compiler probably would complain, at least if someone used `ESAPICustomJavaLevel`.)

However, in my opinion, I think the best thing to do is to just level the old class and wait until 3.0 to rename it. We'll have a whole new package namespace then (I'm thinking "org.owasp.esapi3") so we can then have much freer range of how to name things.

src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
```diff
@@ -18,7 +18,7 @@
 import java.util.logging.Level;
 
 /**
- *  A custom logging level defined between Level.SEVERE and Level.WARNING in logger.
+ *  Definitions of customized Java Logging Level options to map ESAPI behavior to the desired Java Log output behaviors.
```
kwwall, line 21: Probably still advisable to leave this comment here even if the class name is not changed.

src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
kwwall: @jeremiahjstacey -- If you don't think you can get to this for a while, I could just merge this PR and then do a follow-up PR to immediately flip the class names back. Let me know.

jeremiahjstacey: I have restored the ESAPIErrorJavaLevel class from history and marked it as deprecated.
The documentation now suggests using the ESAPICustomJavaLevel.ERROR_LEVEL instead.

To ensure consistency between the ESAPIErrorJavaLevel and the ESAPICustomJavaLevel.ERROR_Level, I have used the values of ESAPICustomJavaLevel.ERROR_LEVEL as updated constructor args for the ESAPIErrorJavaLevel instantiation.

src/main/java/org/owasp/esapi/logging/java/ESAPICustomJavaLevel.java
jeremiahjstacey: This was moved to the ERROR_LEVEL entry, since it is specific to that instance. (line 27-31 of this file)

src/main/java/org/owasp/esapi/logging/java/ESAPIErrorJavaLevel.java
```diff
@@ -27,8 +29,10 @@
     /**
      * Defines a custom error level below SEVERE but above WARNING since this level isn't defined directly
      * by java.util.Logger already.
+     * @deprecated
+     * @see ESAPICustomJavaLevel#ERROR_LEVEL
      */
-    public static final Level ERROR_LEVEL = new ESAPIErrorJavaLevel( "ERROR", Level.SEVERE.intValue() - 1);
+    public static final Level ERROR_LEVEL = new ESAPIErrorJavaLevel( ESAPICustomJavaLevel.ERROR_LEVEL.getName(),ESAPICustomJavaLevel.ERROR_LEVEL.intValue());
```
jeremiahjstacey, line 35: This update ensures that both values remain consistent until this class is deleted.  I verified that the enumeration constructor is using the same values, and this should not impact any existing functionality.

kwwall:
@jeremiahjstacey - Merged your PR, but added on additional commit e0943e2 to get rid of 2 compile-time warnings.

Issue #588, opened 2021-02-02
Possible bug in StringValidationRule.getValid  with canonicalize


DavidSmithAvaloq:
Using version 2.2.1.0
In StringValidationRule.getValid  with canonicalize set I see

```java
		// canonicalize
		if (canonicalizeInput) {
		    data = encoder.canonicalize(input);
		} else {
		    String message = String.format("Input validaiton excludes canonicalization.  Context: %s   Input: %s", context, input);
		    LOGGER.warning(Logger.SECURITY_AUDIT, message);
            data = input;
		}

		// check whitelist patterns
		checkWhitelist(context, input);

		// check blacklist patterns
		checkBlacklist(context, input);
```

This sets data to be the canonical version of the string but then appears to do the validation on the original input string.

xeno6696:
@DavidSmithAvaloq looking at where the method gets called, it's being used in `Validator.getValidInput()`, as well as from `BaseValidationRule`, `CreditCardValidationRule`, and `HTMLValidation` rule.  

That part of the API is designed to assert that the input data is untrusted with an idiomatic approach that suggests the user intends on receiving 'safe' data from an untrusted source.  

@planetlevel -->  Do you by any chance recall conversations as to why we chose to return altered data as opposed to original data that we reasonably assert is safe?  (I.E. --> running canonicalize to detect mixed/multi encoding but ultimately returning the original data?)

planetlevel:
There were some very vocal (if not helpful) members of the community that were loudly against modifying data. I think they thought this would introduce some sort of integrity problem. I disagreed and still disagree. The safe behavior should be the default, and if someone wants to specifically get the unsafe data, it should be possible -- like getOriginalAndPossiblyUnsafeInput() or something.

xeno6696:
@planetlevel  While I think there's room for nuance especially with some aggressive request validation frameworks I won't resurrect old battles and I have a general penchant to NOT change something especially if its as fundamental and longstanding as this bit of code.  

I'm going to close this issue.  For my part @DavidSmithAvaloq I agree with your approach, but given that the philosophy governing this part of the library was built with the opposite intent I'm going to suggest a simple workaround:  If you want to store the original input data, make two calls, one with and one without canonicalize.  You _will_ double the validation workload but given the regexes we use are almost always length-limited this shouldn't be an issue.  Of whitelist/blacklist/canonicalize canonicalize is the most expensive method and would only get called once here.   

[...]

PR #591, opened 2021-02-05
PR to fix #517 and #588

### LINKED ISSUE #517: Encoded input is not treated as an attack
behavior of `getValidInput` method in class `DefaultValidator` has been changed between versions `2.1.0.1` and `2.2.0.0`

**Example** :
`ESAPI.validator().isValidInput("QueryString", "value=%3Cscript%3Ealert(1)%3C%2Fscript%3E", 30000, true)`
**Expected**: return false
**Actual** ( version 2.1.0.1): returns false
**Actual** ( version 2.2.0.0): returns true

xeno6696:
@kwwall @jeremiahjstacey for the moment just try running mvn:test against this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
-    <version>2.2.2.0-SNAPSHOT</version>
+    <version>2.2.2.0</version>
     <packaging>jar</packaging>
 
     <distributionManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
-    <version>2.2.2.0</version>
+    <version>2.2.2.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <distributionManagement>
--- a/src/main/java/org/owasp/esapi/codecs/JavaScriptCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/JavaScriptCodec.java
@@ -211,7 +211,8 @@ public class JavaScriptCodec extends AbstractCharacterCodec {
 		}
 		
 		// ignore the backslash and return the character
-		return second;
+		input.reset();
+		return null;
 	}
 
 }
--- a/src/main/java/org/owasp/esapi/reference/validation/StringValidationRule.java
+++ b/src/main/java/org/owasp/esapi/reference/validation/StringValidationRule.java
@@ -274,16 +274,16 @@ public class StringValidationRule extends BaseValidationRule {
 		if (canonicalizeInput) {
 		    data = encoder.canonicalize(input);
 		} else {
-		    String message = String.format("Input validaiton excludes canonicalization.  Context: %s   Input: %s", context, input);
+		    String message = String.format("Input validation excludes canonicalization.  Context: %s   Input: %s", context, input);
 		    LOGGER.warning(Logger.SECURITY_AUDIT, message);
             data = input;
 		}
 
 		// check whitelist patterns
-		checkWhitelist(context, input);
+		checkWhitelist(context, data, input);
 
 		// check blacklist patterns
-		checkBlacklist(context, input);
+		checkBlacklist(context, data, input);
 			
 		// validation passed
 		return data;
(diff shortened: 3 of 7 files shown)


kwwall:
@xeno6696 - Will look at this tomorrow or Saturday. My brain is fried right now.

kwwall:
@xeno6696 - I take it that for the ESAPI.properties at least, we also want to update the one in 'configuration/esapi/ESAPI.properties' as well? Not sure about the validation.properties one. Some of that looks like it is only for testing.

Could you check and update your PR before tomorrow if possible? AntiSamy 1.6.0 just released today so I want to get a new ESAPI release out over the weekend.

xeno6696:
@kwwall so "mvn test" does not work for me locally on the testLoggerByString and testLoggerByClass methods.  Bear that in mind.  

We have an instance here where I need guidance.  There's a legacy unit test input that fails with the changes needed to clear the two issues in question here.  When the validator runs a check on the input `jeff\WILLIAMS` as an HTTPParameterValue we get a passing unit test when we would expect a failure.  

The JavaScript codec has a possible bug, worth discussing.  If the character sequence we're interested in has a literal `\` like in the affected input, it starts to treat it like an encoded sequence.  However, since javascript encoding is hex, IE `\x###` instead of seeing the next character as `W` and returning the expected value of `\` it resets the pusbhackstring sequence and returns null.  

Thus, after the JavaScriptCodec runs `jeff\WILLIAMS` becomes `jeffWILLIAMS` which will then be the value we use to run input validation against, in this case a regex that doesn't have `\` as a whitelist, thereby the input ought to be rejected but it is accepted.  

This is probably why the original code had that funky "validate the input, THEN canonicalize, THEN validate THAT input again.  

What I haven't yet figured out is precisely what I want to do to resolve this.  It seems to me that I should find a way to finesse the JavaScriptCodec here but I don't know its escaping rules as readily as I do Java or Python--which is why it's been languishing.

xeno6696:
I could be overthinking this.  If it's an HTTPParameterValue then we should be doing a form of canonicalization that is different.  This is like the case that prompted me to create "getCanonicalizedURI()".

xeno6696:
It won't be done today though @kwwall

kwwall:
@xeno6696 - I will look at it.

jeremiahjstacey:
@kwwall , I think @xeno6696  and I were able to isolate a possible fix to the JavaScriptCodec earlier today.

The test case we're trying to prove is:

```
 public void jsEscapeTest() {
     // javascript escape syntax
        ArrayList<String> js = new ArrayList<String>();
        js.add( "JavaScriptCodec" );
        Encoder instance = new DefaultEncoder( js );
        
        assertEquals( "jeff\\WILLIAMS", instance.canonicalize("jeff\\WILLIAMS"));
    }
```
(Please note this syntax is borrowed from the JavaScriptCodec testing done in  `EncoderTest.testCanonicalize()` )

We narrowed the change required down to the `JavaScriptCodec decodeCharacter( PushbackSequence<Character>)` implementation. 

In this case the issue appears to be related to the last return of that method where the first character '\' and the second character 'W' do not represent an expected escape sequence for the Codec implementation.

[JavaScriptCodec](https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/codecs/JavaScriptCodec.java)

```
213      // ignore the backslash and return the character
214      return second;
```  

If we alter the default return logic to acknowledge that the first-found character '\' is  not meant to represent an escaped sequence, and therefore return null (no encoding performed), then the method appears to act as desired.

```
213		input.reset();
214		return null;
```

This change breaks one case in `EncoderTest.testCanonicalize()`:  line 235
[EncoderTest](https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/test/java/org/owasp/esapi/reference/EncoderTest.java)

`235   assertEquals( "<", instance.canonicalize("\\<"));`

Updating that to expect that the backslashes are not stripped matches the updated code and allows all tests to pass.

`235  assertEquals( "\\<", instance.canonicalize("\\<"));`

Last we spoke, I believe that @xeno6696  was going to be verifying this locally and inspecting the logic for other possible implications from this change.

xeno6696:
Analysis work done.  Specified that the test regex in src/test is NOT to be used in production, so I made that clear @kwwall .

This one's ready to ship.

kwwall:
Okay @xeno6696 - thanks for getting on this. And @jeremiahjstacey, thanks for the assist here. I guess it was a wise decision on my part to run out for an errand because when I returned I found you had this resolved. :-)

I will fit the one remaining problem with the ValidationTest that is causing the CI workflow with Maven to fail after I merge Matt's PR.

PR #598, opened 2021-03-03
Update batik-css to address CVE-2020-11987 and CVE-2020-11988

tobiasstadler:


kwwall:
@tobiasstadler - I'm rejecting this PR for few reasons, but I wanted to let you know because I don't want to discourage you from submitting PRs. Here is why I am rejecting merging your PR:

1. I already have both of these in my local changes (along with several others) and committing them and pushing to GitHub this weekend to do a new ESAPI release.
2. The Snykbot already had the batik-css update before you did. (I'm also going to be rejecting it, for reason number 1, above, as well).
3. In general, the ESAPI dev team regularly runs OWASP Dependency Check with makes us aware of this, usually long before Snykbot and the GitHub security bot so in general, we are already covered pretty well with dependency related updates.
4. I have been working with @davewichers for both of these dependency updates. Dave just released AntiSamy 1.6.0 today and I have made changes in anticipation of it.
5. There are actually 3 AntiSamy policy files in the ESAPI source code that had '<html-enties>' node that needed to be removed.
6. A new test needs to be written to have AntiSamy ignore the schema validation of the respective policy file against its XSD. I have written such new JUnit tests.

Again, I appreciate your PR, I really do. But accepting it would mean I'd have to undo a lot of the changes that I have already made.

PR #601, opened 2021-03-06
Fixing system-specific unit test

xeno6696:


xeno6696:
I'm going to close this as the failure I was getting before is different from what I'm seeing now in the CI build.

kwwall:
See issue #602. I think that commit #[936508c](https://github.com/ESAPI/esapi-java-legacy/commit/936508cd34829e8df2ff5aa066558825dff50506) should fix this.

PR #607, opened 2021-03-14
HTMValidationRule & Test Updates

jeremiahjstacey:
Effort to ensure that HTMValidationRuleClasspathTests shows consistent results when run through `mvn clean test` or `mvn clean test -Dtest=HTMLValidationRuleClasspathTest`

Work includes:
  - Splitting out the static classloader method of HTMLValidationRule into smaller static package-protected methods for more isolated interactions
  - Correcting incorrect file being loaded into the HTMLValidationRuleClasspathTest - a fortunate result of the static block refactoring.
  - Updating Antisamy dependency to v 1.6.1
  - Applying work-around for [Antisamy Issue 75](https://github.com/nahsra/antisamy/issues/75) while solution is being implemented.

Currently, a large portion of the static blocks in the HTMLValidationRule do not have targeted test coverage.  The expectation is that this code will be short-lived when the work for pulling the capability back into the DefaultSecurityConfiguration is performed.
#608


jeremiahjstacey:
Failure is pending the updated Antisamy version.  Expected at this stage, and has been communicated with @kwwall 

Release of antisamy 1.6.2 expected within the week, at which point everything here should be good to go.

PR #616, opened 2021-03-27
First public-signed commit request from Matt Seil

xeno6696:
This is to make a public record in the git repository that all **MY** commits in the future must be signed by me in order to be considered authentic.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
-    <version>2.2.4.0-SNAPSHOT</version>
+    <version>2.2.4.1-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <distributionManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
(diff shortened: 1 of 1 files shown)


kwwall:
@xeno6696 - Should I just close this since your signature is now showing up as 'unverified' since you changed the expiration period (and presumably generated a new key pair)? Or do you just want to cancel your PR. Or what? The message is 'No user is associated with the committer email'.)

xeno6696:
I was researching this and I don’t think there’s a safe way to make this work as-is.  I will close this out on my own.

xeno6696:
I can't reproduce this locally =-/

kwwall:
@xeno6696 wrote:
> I can't reproduce this locally =-/

Can't reproduce _what_ locally? The fact that it is unverified? Did you remember to upload your public GPG key to github.com as per https://docs.github.com/en/github/authenticating-to-github/associating-an-email-with-your-gpg-key? (I don't think github.com uses GPG key servers, unlike Sonatype / Maven Central.)

xeno6696:
@kwwall I couldn't get "unverified" to show up locally.  What I wasn't aware of is that github conducts its own check.  I typically set all my email defaults as "xeno6696[at]gmail.com" because I still remember usenet.  Github didn't like that my `git config user.email` wasn't the canonical version of my email address.  

All is fixed now.

kwwall:
Ha! File that under "the lies we tell ourselves". It has always amused me that people think that writing ones email address as '[at]' or '.at.' or ' at ' instead of '@' was somehow sufficient to prevent spammers from harvesting their email address. Uh huh. Yeah. Right. :)

Issue #630, opened 2021-07-16
Incorrrect result for `isEnabled()` in `Slf4JLogger`


sjoshid:
`isEnabled(int esapiLevel)` in `Slf4JLogger` returns incorrect result. Mostly due to the OR condition `maxLogLevel == Logger.ALL`.

## Steps to reproduce (in 2.2.3.1)
* Create a logger for `SomeClass` in log4j.xml. Set its level to `info` level. Like so:
```
<Logger name="com.SomeClass" level="info">
    <appender-ref ref="SOME-APPENDER-REF"/>
</Logger>
```
* `SomeClass` looks like:
```
private static final Logger LOG = ESAPI.getLogger(SomeClass.class);;
...
if(LOG.isDebugEnabled()) {
    // do some heavy operation here. 
}
```

## Expected result
* With above logger config `LOG.isDebugEnabled()` will return `false`. 

## Actual result
* `LOG.isDebugEnabled()` returns `true`.

## Some more observations
* Even though the enabled checks (like `isDebugEnabled()`, `isInfoEnabled()`, etc.) fail, the actual log statements (like `debug()`, `info()`) do honor the log levels. 
* This has a big perf impact because of all the crap that might go in `LOG.isDebugEnabled()` check.

jeremiahjstacey:
I believe this is valid.  Looking at the documentation for `org.owasp.esapi.Logger.setLevel(int)`
``` 
/**
     * Dynamically set the ESAPI logging severity level. All events of this level and higher will be logged from 
     * this point forward for all logs. All events below this level will be discarded.
     * 
     * @param level The level to set the logging level to. 
     */
```

> All events of this level and higher

This is inversed in the loggers, all of them.  Slf4J, Log4J, and JavaLogger implementations.

```
  private boolean isEnabled(int esapiLevel) {
        //Are Logger.OFF and Logger.ALL reversed?  This should be simply the less than or equal to check...
        return (esapiLevel <= maxLogLevel && maxLogLevel != Logger.OFF) || maxLogLevel == Logger.ALL;
    }
```

The bug is the esapiLevel check, not the OR statement.  The code should read
`esapiLevel >= maxLogLevel`

This makes the naming of the variable `maxLogLevel` a bit confusing, which I think is partly the origin of the problem.
I am suggesting the following:

1. Update the `testLevelEnablement` test implementation for each of the loggers.
These tests are currently expecting the inverse logic behavior and will need to be updated to account.
```
  @Test
[...]

PR #632, opened 2021-07-18
Issue 631

### LINKED ISSUE #631: Create Default Logging level configuration for ESAPI library
Derived from issue #630, the default logging level of all ESAPI loggers is `Level.ALL`.
This should be moved to a configurable component in ESAPI.proprites and all of the LoggerFactories should be updated to use that value for generating new Logger instances.

sjoshid:
Fixes #631 
Creating this PR just for discussion. Feel free to close it if you think this is ridiculous.

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogger.java
+++ b/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogger.java
@@ -49,11 +49,9 @@ public class Slf4JLogger implements org.owasp.esapi.Logger {
             logBridge.log(delegate, esapiLevel, type, message, throwable);
         }
     }
-    
 
     private boolean isEnabled(int esapiLevel) {
-        //Are Logger.OFF and Logger.ALL reversed?  This should be simply the less than or equal to check...
-        return (esapiLevel <= maxLogLevel && maxLogLevel != Logger.OFF) || maxLogLevel == Logger.ALL;
+        return logBridge.isEnabled(delegate, esapiLevel);
     }
     
     @Override
--- a/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogBridge.java
+++ b/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogBridge.java
@@ -21,6 +21,7 @@ import org.slf4j.Logger;
  * 
  */
 public interface Slf4JLogBridge {
+    boolean isEnabled(Logger logger, int esapiLevel);
     /**
      * Translation for the provided ESAPI level, type, and message to the specified SLF4J Logger.
      * @param logger Logger to receive the translated message.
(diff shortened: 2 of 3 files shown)


src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogger.java
```diff
@@ -49,11 +49,9 @@ private void log(int esapiLevel, EventType type, String message, Throwable throw
             logBridge.log(delegate, esapiLevel, type, message, throwable);
         }
     }
-    
 
     private boolean isEnabled(int esapiLevel) {
-        //Are Logger.OFF and Logger.ALL reversed?  This should be simply the less than or equal to check...
-        return (esapiLevel <= maxLogLevel && maxLogLevel != Logger.OFF) || maxLogLevel == Logger.ALL;
```
sjoshid, line 56: This also means we dont need `maxLogLevel`

jeremiahjstacey:
Although I agree with the intent of what's being requested, this is not the correct solution.  Although this does fulfill the need of issue #630, it will require rework should we address issue #631.

I do not believe this implementation is sustainable for that reason.  In addition, I'm fairly confident that since there are no test changes in this update that the baseline will currently not pass unit tests.
Please be sure you are running:

`mvn clean test` or `mvn clean package` before submitting PRs -- all tests must pass.

One other thing to note is that whatever change we make with SLF4J must be consistent with Log4J and Java Logger implementations. 

I feel strongly that my suggested solution (https://github.com/ESAPI/esapi-java-legacy/issues/630#issuecomment-881776104) is the most appropriate path given the full context of the discussion.

jeremiahjstacey:
In my initial comment I missed we were trying to address #631. 
I was assuming we were fixing #630, my apologies.

We are not in a position to fix #631 at this time.  We need to correct the logic first, and negotiate the use cases presented by @kwwall  in #631 before we can know that we are making the intended changes.

I agree that the ESAPI library may need to expose configuration between the internal logging and the logging delegation, but at this time we have not had enough discussion to understand what the API contract should be at that boundary.

Although this is one possible approach, I'm not sure it covers all necessary cases -- I'm not really sure I fully understand all of the cases we're concerned with yet.

sjoshid:
This was intended to fix both: #630 and #631.
> mvn clean test or mvn clean package before submitting PRs -- all tests must pass.

Yes. I just wanted to put forth my thoughts. Failing tests need to be changed for sure. One main thing with this PR is there is no change to downstream esapi clients. `maxLogLevel` was never exposed so it would work seamlessly. 

But I understand yours and Kevin concerns about this change. So I'll shut up now. 

Thanks for taking time to go through it.

jeremiahjstacey:
Putting this on hold pending future discussions on #631.

Issue #629, opened 2021-07-11
Define .snyk ignore content


jeremiahjstacey:
SNYK tool continues to report the upgrade of commons-io.  We cannot support that and maintain Java 7 compatibility.  Looking through the SNYK documentation, we should be able to generate a .snyk file and specify an ignore clause and a reason for ignore, which should prevent the tool from flagging the version going forward.
https://github.com/ESAPI/esapi-java-legacy/pull/627
https://snyk.io/vuln/SNYK-JAVA-COMMONSIO-1277109

Based on the SNYK documentation [here](https://support.snyk.io/hc/en-us/articles/360007487097-The-snyk-file), I think the .snyk file should contain approximately what I've outlined below.

```
# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
version: v1.14.0
ignore:
 SNYK-JAVA-COMMONSIO-1277109:
    - commons-io:commons-io:
        reason: ESAPI cannot upgrade past the current commons-io version and still maintain Java 7 compatibility
        expires: '2025-12-30T00:00:00.000Z'
```

I am not entirely sure how to test this content without waiting to see if SNYK flags commons-io on its next pass.

kwwall:
@jeremiahjstacey - The proper way would probably be to configure one of our private repos to and have Snyk test that and create PRs there and then wait. How long? I'm not sure, but it I get a Snyk report every week (but I just don't have it generate PRs any longer, because it was always doing this sort of thing).

But if you and @xeno6696 are okay with it, I say we just create this in the 'develop' GitHub branch and then wait to see what happens. Unless you made a typo in the Snyk identifier, I'm pretty sure the worst that can happen is that we just won't get it ignored.

jeremiahjstacey:
I'm making a branch for this locally where I'm going to address the updates for slf4j and antisamy as well.
https://github.com/ESAPI/esapi-java-legacy/pull/636
https://github.com/ESAPI/esapi-java-legacy/pull/628

kwwall:
Closed via PR #637

PR #637, opened 2021-08-08
Snyk 629

jeremiahjstacey:
This should close issue #629 as well as PR #636  #633 and #628 

The .snyk content is intended to prevent #635 from appearing in the future -- I believe that should also be rejected.


kwwall (approved):
It all looks good to me! And signed commits as well! Woohoo!

PR #638, opened 2021-08-10
Logger Level Logic Updates

### LINKED ISSUE #630: Incorrrect result for `isEnabled()` in `Slf4JLogger`

jeremiahjstacey:
Closes #630 

Tests updated to reflect desired results of `isEnabled` for each of the loggers.
Logic updated to inverse the evaluate such that the method parameter must be greater than or equal to the configured logging level.
The field `maxLogLevel` has been renamed to `loggingLevel`

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogger.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogger.java
@@ -24,7 +24,7 @@ public class JavaLogger implements org.owasp.esapi.Logger {
     /** Handler for translating events from ESAPI context for Java processing.*/
     private final JavaLogBridge logBridge;
     /** Maximum log level that will be forwarded to Java from the ESAPI context.*/
-    private int maxLogLevel;
+    private int loggingLevel;
 
     /**
      * Constructs a new instance. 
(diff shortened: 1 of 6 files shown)


src/main/java/org/owasp/esapi/logging/java/JavaLogger.java
```diff
@@ -52,8 +52,7 @@ private void log(int esapiLevel, EventType type, String message, Throwable throw
 
 
     private boolean isEnabled(int esapiLevel) {
-        //Are Logger.OFF and Logger.ALL reversed?  This should be simply the less than or equal to check...
-        return (esapiLevel <= maxLogLevel && maxLogLevel != Logger.OFF) || maxLogLevel == Logger.ALL;
+        return esapiLevel >= loggingLevel;
```
jeremiahjstacey, line 55: This is the logic update in each file.  All other changes are just the rename event.

PR #643, opened 2021-08-25
Link to release note, Update latest version links

OlivierJaquemet:
- Add direct link to release note of  2.2.3.0 for the AntiSamy warning 
- Latest version is 2.2.3.1, not from 2.2.3.0 update accordingly


kwwall (changes requested):
I'm fine with the 2nd change, since that is obviously something I botched, but that should tell you something... I think I would honestly prefer just removing ALL the links to the release notes as it's just too much of a PITA for long term maintenance to try to keep those updated. For that reason, I'd prefer to NOT add a link to line 37, so please remove that and then I will merge. (Or better, just remove both links. :)

OlivierJaquemet:
Hello @kwwall 

The [first changes line 37](https://github.com/ESAPI/esapi-java-legacy/pull/643/commits/0bf5df9a3ae66078701ad45b2936f170940ad092#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L37) is a link to the 2.2.30 release note, it will never change and the link will always be valid. So there is no maintenance for this link in future releases. 
Therefore, I would personally this exactly like I did in my PR.

The [second changes line 40](https://github.com/ESAPI/esapi-java-legacy/pull/643/commits/0bf5df9a3ae66078701ad45b2936f170940ad092#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L40) is indeed more problematic in the long term as it requires a change for each release. 
I suggest to keep it that way for this PR so it is up to date, and perform the following modification for the long term : 

- in the next release : add the changelog file to the assets, using a constant filename such as changelog.md (you could even do this for existing releases, cf https://docs.github.com/en/rest/reference/repos#upload-a-release-asset)
- in the readme.md : update the links line 40 to point to the constant path ESAPI/esapi-java-legacy/releases/latest/assets/changelog.md, and remove any mention of the latest version number

What do you think ?

kwwall (approved):

Issue #661, opened 2022-02-23
add OSGI metadata


zspitzer:
there's a maven plugin which adds OSGI metadata

because it's missing, there are quite a few projects which wrap the library to add headers, this causes security scanners to miss flagging outdated versions

kwwall:
I am preparing a release that, among other updates, will add an SBOM via Steve Springett's [cyclonedx-maven-plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin) that produces a SBOM in XML and JSON formats that is compatible with [OWASP Dependency Track](https://owasp.org/www-project-dependency-track/). Examples (renamed to .txt so I could attach them) from that  plug-in is attached.

Would this SBOM not serve the same purpose.?The ESAPI pom is already way more complicated that I'd like it to be and it seems like I spend 1/4 of my life trying to keep all the plugins updated.
[bom.xml.txt](https://github.com/ESAPI/esapi-java-legacy/files/8126004/bom.xml.txt)
[bom.json.txt](https://github.com/ESAPI/esapi-java-legacy/files/8126007/bom.json.txt)

xeno6696:
I would think the OWASP Dependency-Track covers this.  I was taking a peek into this one in particular, and the OSGi framework is bloated IMHO.  

I can see the value-add, but OSGi has been around since '99 and you'd think this would have bubbled up in importance if this was something that was under heavy adoption.  

I would think that the OWASP Dependency Track would be sufficient.  I poked around in this, and to implement OSGi metadata requires adding a custom class to ESAPI to instruct the build operations.  I only looked into it for a few minutes, but this seems like overkill if all you need is a manifest file listing dependencies.

xeno6696:
Nevermind @kwwall this is easy.  I figured out that you can leave your build package alone and just generate the metadata file.  

`mvn org.apache.felix:maven-bundle-plugin:manifest`

Before a standard "mvn package" command will generate the file, package will automatically scoop it up into META-INF/

kwwall:
@xeno6696 - Easy, perhaps. That wasn't so much of my concern. But is it _necessary_ if we are creating an SBOM? It complicates the pom which I don't really like. The SBOM contains all the same info as near as I can tell, along with hashes and I've made changes suggested by Dave Wichers to introduce reproducible builds as well.

kwwall:
[...]

PR #663, opened 2022-02-25
Housekeeping, OSGI start/stop, Parameterized some hardcoded values.  

xeno6696:
Also, this updates AntiSamy to 1.6.5.  This also records the changes for #661 and the revert commit.  (If we ever want to return to OSGI manifests it's at least in the revision history.)


configuration/esapi/ESAPI.properties
```diff
@@ -469,7 +469,7 @@ Validator.Redirect=^\\/test.*$
 Validator.HTTPScheme=^(http|https)$
 Validator.HTTPServerName=^[a-zA-Z0-9_.\\-]*$
 Validator.HTTPCookieName=^[a-zA-Z0-9\\-_]{1,32}$
-Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
+Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]{1,1024}$
```
kwwall: I thought you had planed to change this regex to a range of `{0,1024}` to allow for cookies with no value? Did that not work?

src/test/resources/esapi/ESAPI.properties
```diff
@@ -498,7 +498,7 @@ Validator.Redirect=^\\/test.*$
 Validator.HTTPScheme=^(http|https)$
 Validator.HTTPServerName=^[a-zA-Z0-9_.\\-]*$
 Validator.HTTPCookieName=^[a-zA-Z0-9\\-_]{1,32}$
-Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
+Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]{1,1024}$
```
kwwall: Similar comment: I thought you had planed to change these regexes to a range of `{0,1024}` to allow for cookies with no value? Did that not work?

Ideally we should make these changes too ALL the ESAPI properties files under src/test/resources so they don't diverge over time.

kwwall (changes requested):

xeno6696 (commented):
@kwwall the change to zero is in this commit here.

src/test/resources/esapi/ESAPI.properties
xeno6696: see commit [5246396](https://github.com/ESAPI/esapi-java-legacy/pull/663/commits/5246396ee4b7a2aa77c4f1fa20375721e57ee02d)

kwwall: Okay, but can you also change this to be like the others in **ESAPI.properties**?
```
src/test/resources/esapi/ESAPI-CommaValidatorFileChecker.properties:471:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
src/test/resources/esapi/ESAPI-DualValidatorFileChecker.properties:472:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
src/test/resources/esapi/ESAPI-QuotedValidatorFileChecker.properties:470:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
src/test/resources/esapi/ESAPI-SingleValidatorFileChecker.properties:470:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
```
I'd prefer to keep them as similar as possible since when then diverge a lot there is a lot of diffing that needs to be done to get them sync'd back up. Otherwise, we should strip down those other for properties files to them minimal set used to make the tests succeed.

xeno6696: Maybe I messed this up, but I fixed it in src/main before src/test.  (Just what I had open.)

I’ll double check when I get upstairs

xeno6696: It used to be src/test and src/main but I thought src/main was replaced with configuration/

kwwall: No, you didn't mess it up. I just asked that you make the same change to the other 4 properties files in 'src/test/resources'.
```
src/test/resources/esapi/ESAPI-CommaValidatorFileChecker.properties:471:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
src/test/resources/esapi/ESAPI-DualValidatorFileChecker.properties:472:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
src/test/resources/esapi/ESAPI-QuotedValidatorFileChecker.properties:470:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
src/test/resources/esapi/ESAPI-SingleValidatorFileChecker.properties:470:Validator.HTTPCookieValue=^[a-zA-Z0-9\\-\\/+=_ ]*$
```

Everyone forgets to update them. Including myself. The need to have the regex changed so that the range check `{0,1024}` is added at the end instead of the '*' at the end.

PR #676, opened 2022-04-03
Upgrade/cleanup pom and prepare a bit for AntiSamy 1.6.6.1

davewichers:
Upgrade a bunch of dependencies and plugins and prepare a bit for the use of AntiSamy 1.6.6.1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,11 +132,13 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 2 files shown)


xeno6696:
Thanks for the help @davewichers especially for parameterizing it all.  

To answer your question from the email, I normally run mvn site -DskipTests 

ESAPI wasn't the first project I've used that had that weird little quirk.  I'm not sure what the issue is.  I studied the reference quick and you're right that site looks like its supposed to be self-contained.  I'll have to research this.

Issue #679, opened 2022-04-10
Completely remove support for fixed IVs and throw a ConfigurationException if encountered.


kwwall:
In early 2.0 and 2.1 versions of ESAPI, fixed IVs  were supported, but that turned out to be really bad idea. Thus, way back in release 2.2.0.0, this feature was deprecated and if it was used, a prominent warning (which probably everyone ignored) was displayed if you used it (that is, if you had the ESAPI property '**Encryptor.ChooseIVMethod**' set to '**fixed**'.) Unfortunately, knowing human nature, those who were using it probably just ignored the warning message.

Consequently, it is now time to completely disable it and change ESAPI from just displaying a warning to instead throwing a `ConfigurationException`.

This issue should actually remove the old code that implemented this.

kwwall:
Closed as per PR #680. This will first appear in the ESAPI 2.3.0.0 release.

PR #680, opened 2022-04-11
Changes to close issue #679

kwwall:
@xeno6696 or @jeremiahjstacey -- these changes are not really as much crypto-related (other than the _reason_ why I'm making them), as the are related to the ESAPI `SecurityConfiguration` component.

I originally had them in the otherwise small updates to prep the next release (which I am actually going to call 2.3.0.0 rather than 2.2.4.0, so I can finally justify dumping this whole bad "fixed IV" idea. (Thought up in the days where I was emphasizing backward compatibility and supporting legacy code over security, Sigh. Well, guess that means I learned _something_ in the past 10 plus years.)

Anyhow, could I get one of you to do a quick and dirty code review and them merge this? Pending my email to Sebastian and Dave, I'd like to get this merged by early tomorrow evening so I can get it into the release. Thanks!

P.S.- I assigned _both_ of you as code reviewers, but that doesn't mean that you both really need to look at it. There is actually a LOT more cleanup that I can do, but I don't want to shock people too much and remove functionality that would actually keep code from compiling, which is why I've left the deprecated SecurityConfiguration.getIVType() for now, but it's pointless now that I've changed the code to only allow a single valid value.

### DIFF
--- a/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
@@ -121,10 +121,9 @@ public class DefaultSecurityConfiguration implements SecurityConfiguration {
     public static final String CIPHER_TRANSFORMATION_IMPLEMENTATION = "Encryptor.CipherTransformation";
     public static final String CIPHERTEXT_USE_MAC = "Encryptor.CipherText.useMAC";
     public static final String PLAINTEXT_OVERWRITE = "Encryptor.PlainText.overwrite";
-    public static final String IV_TYPE = "Encryptor.ChooseIVMethod";
 
     @Deprecated
-    public static final String FIXED_IV = "Encryptor.fixedIV";
+    public static final String IV_TYPE = "Encryptor.ChooseIVMethod";    // Will be removed in future release.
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
```diff
@@ -832,49 +838,26 @@ public boolean overwritePlainText() {
...
+    		throw new ConfigurationException("Contrary to previous internal comments, '" + IV_TYPE + "=specified' is not going to be supported -- ever.");
     	} else {
-    		// DISCUSS: Should we just log a warning here and return null instead?
-    		//			If so, may cause NullPointException somewhere later.
-    		throw new ConfigurationException("IV type not 'fixed' [which is DEPRECATED!] (set to '" +
-    										 getIVType() + "'), so no fixed IV applicable.");
+    		logSpecial("WARNING: '" + value + "' is illegal value for " + IV_TYPE +
+    										 ". Using 'random' for the IV type.");
     	}
+        return "random";
```
jeremiahjstacey, line 860: Minor.  If the logic falls through here the `value` parameter is still set to the default 'random'

Could update `return value;`

PR #688, opened 2022-05-03
Issues 686 & 689 - Create DefaultEncoder from Encoder.DefaultCodecList and Javadoc clean-up

### LINKED ISSUE #686: ESAPI canonicalization in DefaultEncoder ignoring Encoder.DefaultCodecList property
In the **ESAPI.properties** file, there is this comment and property name / value:
```
# The default list of codecs to apply when canonicalizing untrusted data. The list should include the codecs
# for all downstream interpreters or decoders. For example, if the data is likely to end up in a URL, HTML, or
# inside JavaScript, then the list of codecs below is appropriate. The order of the list is not terribly important.
Encoder.DefaultCodecList=HTMLEntityCodec,PercentCodec,JavaScriptCodec
```
However, contrary to the comment here, this property seems to be getting ignored by the private default CTOR for `DefaultEncoder` and the `HTMLEntityCodec`, `PercentCodec`, and `JavaScriptCodec` are loaded in that order via being hard-coded. However, there is a 2nd public CTOR that takes a `List<String>` representing available (or customized) codecs that could be used with this property. (Question: When is a singleton not a singleton? Answer: When someone uses the other public CTOR. Poor design, but fortunately not any mutable instance variables.)

So, obviously a developer use the ESAPI property with the public CTOR using something like:

```
import org.owasp.esapi.Encoder;
import org.owasp.esapi.reference.DefaultEncoder;

    ...
    Encoder enc =  DefaultEncoder( ESAPI.securityConfiguration().getDefaultCanoicalizationCodecs() );

    String canonicalized = enc.canonicalize( input );
    ...
```

But I don't think that is what was implied by the comment in ESAPI.properties, nor do I think that we want to encourage the direct invocation of a CTOR in a reference implementation class that was intended to be a singleton instance in the first place. So many things wrong with that.

### LINKED ISSUE #689: Clean-up ESAPI Javadoc
- Eliminate all the javadoc warnings shown by '`mvn javadoc:javadoc`'
- Fix some of the missing or incorrectly rendered Javadoc associated with ESAPI `Encoder`, `DefaultEncoder`, and various `Codec` classes.

kwwall:
Close #686. Close #689
LOTS of Javadoc clean-up and corrections.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -439,7 +439,7 @@
             <plugin>
                 <groupId>org.cyclonedx</groupId>
                 <artifactId>cyclonedx-maven-plugin</artifactId>
-                <version>2.5.3</version>
+                <version>2.6.1</version>
                 <executions>
                   <execution>
                     <phase>package</phase>
@@ -787,6 +787,8 @@
                 <configuration>
                     <doclint>none</doclint>
                     <source>8</source>
+                    <failOnError>true</failOnError>
+                    <failOnWarnings>true</failOnWarnings>
                 </configuration>
             </plugin>
             <plugin>
--- a/src/main/java/org/owasp/esapi/Encoder.java
+++ b/src/main/java/org/owasp/esapi/Encoder.java
@@ -160,11 +160,20 @@ public interface Encoder {
     /**
      * This method is equivalent to calling {@code Encoder.canonicalize(input, restrictMultiple, restrictMixed);}.
      *
-     * The default values for restrictMultiple and restrictMixed come from {@code ESAPI.properties}
+     * The <i>default</i> values for {@code restrictMultiple} and {@code restrictMixed} come from {@code ESAPI.properties}.
      * <pre>
      * Encoder.AllowMultipleEncoding=false
      * Encoder.AllowMixedEncoding=false
      * </pre>
+     * and the default codecs that are used for canonicalization are the list
+     * of codecs that comes from:
+     * <pre>
+     * Encoder.DefaultCodecList=HTMLEntityCodec,PercentCodec,JavaScriptCodec
+     * </pre>
+     * (If the {@code Encoder.DefaultCodecList} property is null or not set,
+     * these same codecs are listed in the same order. Note that you may supply
+     * your own codec by using a fully cqualified class name of a class that
+     * implements {@code org.owasp.esapi.codecs.Codec<T>}.
      *
      * @see #canonicalize(String, boolean, boolean)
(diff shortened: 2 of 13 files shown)


kwwall:
@xeno6696 and @jeremiahjstacey - Most of these commits are strictly Javadoc changes. I will leave it up to you whether or not you want to "code review" those changes, since strictly speaking, it isn't code. The main code change was for the `DefaultEncoder` no-args CTOR and the `EncoderTest`. In particular, it would to look closely at the new `testGetDefaultCanonicalizationCodecs()` test. I tried to think of a better 'testInput' string, but everything thing that I could think of that would be flagged by the `UnixCodec` was also being flagged by the `PercentCodec`. I'm sure someone could do better so feel free to suggest major revisions there as long as it doesn't require rewriting all the other tests.

jeremiahjstacey:
Went through the changes.  I agree this content addresses both listed issues 686 and 689.
@xeno6696, I think this is fine to merge pending your review.

jeremiahjstacey (approved):

xeno6696:
I still have to look over the test he was talking about.  Not sandbagging, just crazy busy.

kwwall (commented):
@jeremiahjstacey - Sorry to springing this on you _after_ you already completed the code review, but there's only 3 lines changed here if you want to look at this last commit. The updates to the `maven-javadoc-plugin` are in keeping with issue #689 and the CycloneDX plugin just bumped from2.6.0 (which was in a recent PR) to 2.6.1 so figured I should address those as well. Making 'mvn site' fail on Javadoc warnings and errors should cause us to pay a bit more attention to it.

kwwall:
> I still have to look over the test he was talking about. Not sandbagging, just crazy busy.

@xeno6696 - Matt, I don't think the test is a showstopper. I think it shows that the change works as intended, but it isn't the most elegant test and I think we probably could do better. I was sort of stuck on using the same standard 3 codecs we've always been using (`HTMLEntityCodec`, `PercentCodec`, and `JavaScriptCodec`) and then wanted to write a test where the initial (default) case would pass without an `IntrusionException` but when I swapped out the `JavaScriptCodec` for the `UnixCodec` it would actually throw an `IntrusionException`. But since both `JavaScriptCodec` and `UnixCodec` both do backslash encoding, everything that triggered mixed encoding for the `UnixCodec` also triggered it for the `JavaScriptCodec`. Maybe I should have used one of the DB codecs such as `MySQLCodec` instead. I admittedly just got tired and gave up on the elegant test. Don't let the perfect become the enemy of the good, right?

PR #712, opened 2022-06-28
Add forward slash encoding to DefaultEncoder's encodeForLDAP and encodeForDN

noloader:
According to [1] and [2], the forward slash ('/') character should be encoded for LDAP filters and distinguished names.

[1] https://docs.microsoft.com/en-us/windows/win32/adsi/search-filter-syntax
[2] https://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx

### DIFF
--- a/src/test/java/org/owasp/esapi/reference/EncoderTest.java
+++ b/src/test/java/org/owasp/esapi/reference/EncoderTest.java
@@ -548,7 +548,7 @@ public class EncoderTest extends TestCase {
         assertEquals("No special characters to escape", "Hi This is a test #��", instance.encodeForLDAP("Hi This is a test #��", false));
         assertEquals("Zeros", "Hi \\00", instance.encodeForLDAP("Hi \u0000", false));
         assertEquals("LDAP Christams Tree", "Hi \\28This\\29 = is * a \\5c test # � � �", instance.encodeForLDAP("Hi (This) = is * a \\ test # � � �", false));
-        assertEquals("Forward slash for \\/Microsoft\\/ \\/AD\\/", instance.encodeForLDAP("Forward slash for /Microsoft/ /AD/"));
+        assertEquals("Forward slash for \\2fMicrosoft\\2f \\2fAD\\2f", instance.encodeForLDAP("Forward slash for /Microsoft/ /AD/"));
     }
     
     /**
@@ -565,7 +565,7 @@ public class EncoderTest extends TestCase {
         assertEquals("less than greater than", "Hello\\<\\>", instance.encodeForDN("Hello<>"));
         assertEquals("only 3 spaces", "\\  \\ ", instance.encodeForDN("   "));
         assertEquals("Christmas Tree DN", "\\ Hello\\\\ \\+ \\, \\\"World\\\" \\;\\ ", instance.encodeForDN(" Hello\\ + , \"World\" ; "));
-        assertEquals("Forward slash for \\2fMicrosoft\\2f \\2fAD\\2f", instance.encodeForDN("Forward slash for /Microsoft/ /AD/"));
+        assertEquals("Forward slash for \\/Microsoft\\/ \\/AD\\/", instance.encodeForDN("Forward slash for /Microsoft/ /AD/"));
     }
     
     /**


kwwall:
When we did this code originally, I think we only based it off RFC 1823,
which is old and probably superseded by something newer.  At the time,  I
think it only supported LDAPv2 though. I will look into it though. Thanks.

-kevin

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -349,12 +354,18 @@ public String encodeForDN(String input) {
         if ((input.length() > 0) && ((input.charAt(0) == ' ') || (input.charAt(0) == '#'))) {
             sb.append('\\'); // add the leading backslash if needed
         }
+        // According to [1] and [2], the encoder should escape forward slash ('/') in DNs.
```
kwwall, line 357: @noloader - Not a big deal, but we probably don't need this same comment in here multiple times. I'd recommend removing this one. The comment on lines 308-310 should be sufficient.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -305,13 +305,18 @@ public String encodeForLDAP(String input, boolean encodeWildcards) {
         }
         // TODO: replace with LDAP codec
         StringBuilder sb = new StringBuilder();
+        // According to "Special Characters" at [1], the encoder should escape '*', '(', ')', '\', '/', NUL. Also see [2].
```
kwwall, line 308: @noloader - In reference [1], that Microsoft documentation definitely lists '/' as requiring encoding, but that reference also references [RFC 2254](https://datatracker.ietf.org/doc/html/rfc2254), and if you look at Section 4 of that document (which defines search filters for LDAPv3), you will not find '/' as being mentioned as requiring encoding.  So, I am thinking this is something that is Active Directory specific. (Thanks Microsoft! Another standard ignored.)

Consequently, I'm not really sure what to do about this. It seems important in that AD is one of the most common LDAP interfaces that people deal with, but if we add we make encoding '/' the default, which I'm not really for, it could break other pure LDAPv3 search filters that conform specifically to RFC 2254.

I guess one ugly option would be to define and document an new `Encoder` interface method that looks like:
```
    public String encodeForLDAP(String input, boolean encodeWildcards, boolean isActiveDirectory)
```
and have that last argument be treated by the other 2 `encodeForLDAP()` methods as the `isActiveDirectory` flag being set to `false`.

That's not perfect, but at least it would keep from breaking things with non-AD LDAPv3 compliant directory servers. The only other (and a way more kludgy mechanism IMO) alternative that I see would be to make some new property in **ESAPI.properties**, such as
```
    Encoder.LDAP.isActiveDirectory=false
```
We'd have it default to `false` so that it wouldn't change existing behavior.

@xeno6696 and @jeremiahjstacey - What are your thoughts on this?

kwwall:
Created Issue #713 to address this.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
noloader: > [RFC 2254](https://datatracker.ietf.org/doc/html/rfc2254), and if you look at Section 4 of that document (which defines search filters for LDAPv3), you will not find '/' as being mentioned as requiring encoding. So, I am thinking this is something that is Active Directory specific. (Thanks Microsoft! Another standard ignored.)

I was thinking... there are three well known X.500 directories: Microsoft's Active Directory, Novell's eDirectory, and U&L LDAP which is based on the RFCs. Microsoft requires forward slash escaping, but the other two may consume an escaped forward slash.

It looks like [RFC 4517](https://datatracker.ietf.org/doc/html/rfc4517) specifies an escaped forward slash in its grammar:

```
PrintableCharacter = ALPHA / DIGIT / SQUOTE / LPAREN / RPAREN /
                           PLUS / COMMA / HYPHEN / DOT / EQUALS /
                           SLASH / COLON / QUESTION / SPACE
      PrintableString    = 1*PrintableCharacter
      IA5String          = *(%x00-7F)
      SLASH              = %x2F  ; forward slash ("/")
      COLON              = %x3A  ; colon (":")
      QUESTION           = %x3F  ; question mark ("?")
```
I'm wondering if the other Directories will just accept the encoded version of the forward slash so you don't need `boolean isActiveDirectory`.

noloader: I asked if OpenLDAP handles the escaped forward slash as expected on the [openldap-technical mailing list](https://lists.openldap.org/hyperkitty/list/openldap-technical@openldap.org/2022/6/). The message is being held for moderation, but it should post soon.

xeno6696: @kwwall I would say that this is what the Codecs were designed for.  Fork the LDAP Codec, and create one for each desired encoding/decoding instance.  Implementers that have to use specific LDAP codecs would then have that choice and could set them in `ESAPI.properties`

Unless I'm missing something this is a quick fix.

noloader:
Hi Everyone/@kwwall,

It looks like the OpenLDAP folks weighed in at [Filter string with escaped forward slash](https://lists.openldap.org/hyperkitty/list/openldap-technical@openldap.org/thread/3QPDDLO356ONSJM3JUKD7NMPOOIKIQ5T/). According to the OpenLDAP folks, [RFC 4515, Section 3](https://tools.ietf.org/search/rfc4515#section-3) says it is Ok to escape forward slash (and other characters).

Given it is Ok to escape the character under all LDAP implementations per the RFC (and required on MS AD), I would merge the PR or something similar.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
noloader: v2 removes the second comment.

noloader:
Hi Everyone/@kwwall,

> please at a new test specifically for this case in **src/test/java/org/owasp/esapi/reference/EncoderTest.java**.

Done in v3.

kwwall:
@noloader - Looks like your new test is failing:

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 s - in org.owasp.esapi.errors.EnterpriseSecurityExceptionTest
[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    EncoderTest.testEncodeForDN:568 expected:<Forward slash for \[2fMicrosoft\2f \2fAD\2f]> but was:<Forward slash for \[/Microsoft\/ \/AD\/]>
[INFO] 
Error:  Tests run: 4326, Failures: 1, Errors: 0, Skipped: 0
```

noloader:
> @noloader - Looks like your new test is failing:
> 
> ```
> [INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 s - in org.owasp.esapi.errors.EnterpriseSecurityExceptionTest
> [INFO] 
> [INFO] Results:
> [INFO] 
> Error:  Failures: 
> Error:    EncoderTest.testEncodeForDN:568 expected:<Forward slash for \[2fMicrosoft\2f \2fAD\2f]> but was:<Forward slash for \[/Microsoft\/ \/AD\/]>
> [INFO] 
> Error:  Tests run: 4326, Failures: 1, Errors: 0, Skipped: 0
> ```

Whoops, copy/paste + failure to save got me. Fixed in v4.

kwwall:
Closes issue #713

PR #714, opened 2022-07-01
Log4J Removal

jeremiahjstacey:
Removal of the Log4J dependency and related implementations.

Last commit updates to the latest versions for a few plugins and the spotbugs dependency.  I was able to run the site w/out issue, so I think that should be fine.


src/main/java/org/owasp/esapi/waf/ESAPIWebApplicationFirewallFilter.java
```diff
@@ -60,448 +58,436 @@
...
+	public void setConfiguration(String policyFilePath, String webRootDir) throws FileNotFoundException {
+
+		FileInputStream inputStream = null;
+
+		try {
+			inputStream = new FileInputStream(new File(policyFilePath));
+			appGuardConfig = ConfigurationParser.readConfigurationFile(inputStream, webRootDir);
+			lastConfigReadTime = System.currentTimeMillis();
+			configurationFilename = policyFilePath;
+		} catch (ConfigurationException e) {
```
kwwall, line 102: If we're touching this line, we might as well fix the typo. Should be 'threw' rather than 'through'. (Probably my fault. I do that all the time and it sounds like one of my comments.)

src/main/java/org/owasp/esapi/waf/ESAPIWebApplicationFirewallFilter.java
```diff
@@ -60,448 +58,436 @@
...
-         * object.
-         */
-
-        FileInputStream inputStream = null;
-
-        try {
-            String webRootDir = fc.getServletContext().getRealPath("/");
-            inputStream = new FileInputStream(configurationFilename);
-            appGuardConfig = ConfigurationParser.readConfigurationFile(inputStream, webRootDir);
-            DOMConfigurator.configure(realLogSettingsFilename);
```
kwwall, line 195: What the heck was the point of this call, anyway? Just to see if the log settings file was a legitimate log4j.xml type configuration file? Or was it expected that since originally ESAPI was using Log4J as it's default logger, tat this just allowed the ESAPI WAF to log to a different file than the rest of ESAPI?

I know that the WAF unit tests all pass without this call to DOMConfigurator, but OTOH I don't have warm fuzzies about the thoroughness of its tests.

kwwall (approved):
@jeremiahjstacey  - Just two comments to consider. One is fixing a typo and the other is a question. Approving this even if you don't address either. If I don't hear from you by Tues, 7/5, I will go ahead and merge. Sure will be nice to get right of this many files though. No bugs to fix in deleted files.

src/main/java/org/owasp/esapi/waf/ESAPIWebApplicationFirewallFilter.java
jeremiahjstacey: Easy enough.  This has been changed.

src/main/java/org/owasp/esapi/waf/ESAPIWebApplicationFirewallFilter.java
jeremiahjstacey: I think it's assuming that WAF has some specific settings that are required, and that those settings cannot be added directly to a pre-existing configuration.
I'm not sure of the use case exactly.

I'm not really sure what to do with this but to dissect the WAF test suites and try to make sense of it.  I can look at this, but I have a feeling it will take a bit to get my head around what's going on.

kwwall: Hey @nahsra  - Can you give us any insight into the ESAPI WAF use case here -- of what this `Log4J DOMConfigurator` was used for? I know if was a long time ago, but you were the original author, so I'm hoping you remember because we are trying to remove the dependency on Log4J 1. here. But OTOH, I don't want to break anything for any clients who might be making advanced use of the WAF and I'm not fully confident that the WAF unit tests are complete enough to test this particular unknown (to us) functionality.

nahsra: If I had to guess, it had something to do with the WAF originally being as "drop-in" utility that wanted to operate on its own logging settings, rather than rely on the app it was included in. 

I don't see any harm in nuking it and letting the WAF portion use ESAPIs typical logging strategy.

kwwall: Okay, thanks for taking the time to look at it Arshan.

PR #716, opened 2022-07-12
Log4j loose ends

kwwall:
In preparation for the ESAPI 2.5.0.0 release, which primarily addresses the removal of the Log4J 1 dependency for ESAPI logging (which has been deprecated for almost 2 years), this PR does the following:
1. Removes the final (non-documentation) vestiges of Log4J. Mostly was left in examples under 'src/examples', scripts under 'scripts' and in various comments. I did not remove it from documentation because of historical needs (e.g., people needing to look at old release notes or security bulletins).
2. Update pom.xml to use latest dependencies and Maven plugins. (Note: The official release of AntiSamy 1.7.0 is not yet in Maven Central, but we will upgrade to that once it is.)

Note that this PR does NOT include the release notes for 2.5.0.0.

### DIFF
--- a/documentation/LoggerDesignAndTesting.md
+++ b/documentation/LoggerDesignAndTesting.md
@@ -22,6 +22,6 @@ The general workflow is:
 
     Logger.info/warn/etc(message) -> forwards to LogBridgelog(logger, esapiLevel, type, message) -> forwards to LogHandler.log(...) -> forwards to slf4j Logger implementation with appropriate level and composed message.
 
-So each of the tests for each of the classes verifies data in -> data out based on the Logging API.  The structure for JUL, Log4J, and SLF4J are almost identical.  There are a few differences in the interaction with the underlying Logger interactions and expectations.  As a result, the tests are also almost full duplications (again accounting for differences in the underlying logging API).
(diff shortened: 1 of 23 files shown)


kwwall:
@jeremiahjstacey - This code review should go really fast. Almost all the changes are for deletions. The only one that you may want to look at ishttps://github.com/ESAPI/esapi-java-legacy/pull/716/files#diff-bbe0e91a04c7ef59cd98c496d3bb3672fd6c3a70b0fe1df51ee7958e79d838ca and to take a quick glance at src/test/java/org/owasp/esapi/waf/WAFTestUtility.java and my changes there and see if you can think of any _simple_ means to test the log output.

jeremiahjstacey:
@kwwall, there are 2 ways that I can think of that may be achievable.  I do not think they're worth the effort, but I've outlined them below as options for consideration:

1.  Update logger configuration for this test to write to the console, then override the System.out stream and capture/compare output.
2. Update logger configuration to write to a file.  At the end of the test read the file and check for output.

In either case, checking for output can be something as direct as verifying the WAF class reference exists.

kwwall:
@jeremiahjstacey or @xeno6696 - Would one of you kindly merge? That looks better than me merging my own code. Thanks.

PR #722, opened 2022-07-17
JSON string escaping support

noloader:
This commit adds JSON string escaping support per RFC 8259, Section 7.

The encoder escapes special characters using two-character escape sequences for popular characters (like \r, \n and \t). For other characters which require escaping, the six-character representation is used (like \u0000).


noloader:
Hi Everyone/@kwwall,

We need this encoder at $dayjob.

kwwall:
I will take a look,  but no promises that  I can get this into the 2.5.0.0
release,  which I am trying to get out today.

Also, most devs that I've seen just use the GSON library for this use case
because they are already using it for most other JSON related use cases.

-kevin

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1050,4 +1051,176 @@ public void testGetDefaultCanonicalizationCodecs() {
 
         assertEquals( true, caughtExpected );   // Verify it threw an IntrusionException
     }
+
+    /**
+     * Test of encodeForJSON method, of class org.owasp.esapi.Encoder.
+     */
+    public void testEncodeForJSON() {
```
xeno6696, line 1058: I've never gotten around to rewriting the tests in here, but in the future I'd like to enforce the idea that we limit 1 test to 1 test method.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1050,4 +1051,176 @@ public void testGetDefaultCanonicalizationCodecs() {
...
+        assertEquals( "\\n", instance.encodeForJSON("\n") );
+        assertEquals( "\\t", instance.encodeForJSON("\t") );
+        assertEquals( "\\\"", instance.encodeForJSON("\"") );
+        assertEquals( "\\/",  instance.encodeForJSON("/" ) );
+        assertEquals( "\\\\", instance.encodeForJSON("\\") );
+
+        // All Unicode characters may be placed within the quotation marks,
+        // except for the characters that MUST be escaped: quotation mark,
+        // reverse solidus, and the control characters (U+0000 through U+001F).
+        for( int i=0; i<=0x1f; ++i ) {
```
xeno6696, line 1088: Very concise... I really like this.  First thing I was checking for was if all characters were being tested.  

We MAY run into downstream issues from the fact that the choice of AbstractCharacterCodec instead of AbstractIntegerCodec will cause UTF-8 encoded characters to be destroyed.  

What happens when the character is 😂, otherwise known as U+1F602?  

Could you add a unit test that covers this case at least, and that we we have a documented test case in code at least that records a known failure? (I suspect that 😂 will be downcast into gibberish)

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1050,4 +1051,176 @@ public void testGetDefaultCanonicalizationCodecs() {
...
+        // And the remainder of printable characters
+        for( int i=32; i<=126; ++i ) {
+            final String str = String.format( "\\u%04x", i );
+            final Character ch = (char)i;
+
+            assertEquals( Character.toString(ch), instance.decodeFromJSON(str) );
+        }
+
+        // And a couple extra for good measure...
+        assertEquals( "\\", instance.decodeFromJSON("\\u005c") );
```
xeno6696, line 1160: No need to change it now, but this is an example of a test case I would want to see isolated into its own `public void testDoubleEncoded()`

noloader:
@kwwall,

> I will take a look,  but no promises that  I can get this into the 2.5.0.0
> release,  which I am trying to get out today.

No problems. The release is more important.

The encoder is available if you want it. If not, we can carry patches for it now that it is written.

xeno6696 (changes requested):
I just want to see a unit test that covers the case `😂, otherwise known as U+1F602` as I want to document clearly what happens when the codec encounters characters that do not exist in the range of legal `Character` or `char` items.  

Ideally  I'd like to see this altered to use the `AbstractIntegerCodec` if that unit test fails like I think it will.  But as this is the first substantial contribution for a Codec since I started working on this project any new functionality is welcome!  

I will also request some unit tests to accommodate mixed encoding behavior, for example, encode a string that contains HTML entities and then do an encoding pass with the JSON Encoder to view what the intermediate result looks like. 

Usually it's corner cases like this where we're likely to see unexpected failures when the Codec is used in chains.

xeno6696:
> I will take a look, but no promises that I can get this into the 2.5.0.0 release, which I am trying to get out today. Also, most devs that I've seen just use the GSON library for this use case because they are already using it for most other JSON related use cases.
> […](#)
> -kevin

@kwwall It's not a make/break feature as in the past I've just had people do javascript escapes for JSON payload data.  It's overkill but it works.  This one is lighter weight.

noloader:
@xeno6696,

>  I'd like to see this altered to use the `AbstractIntegerCodec` ...

Done. I rebased and pushed an updated version.

I'm still not convinced this codec handles all cases well. I'm going to get some test cases added for `U+1F602` and friends.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
noloader: Done. All the encode and decode tests were broken out into their constituent parts.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
noloader: Done. The monolithic self tests were broken-out into their constituent parts.

noloader:
Hi Everyone,

One question I have.... I used `IllegalArgumentException` for an encoding error when decoding. Should this use an ESAPI-specific exception?

src/test/java/org/owasp/esapi/reference/EncoderTest.java
noloader: > What happens when the character is joy, otherwise known as U+1F602?

I checked-in a test case for this. The codec can round-trip it through the encoder and decoder.

I'm still not convinced things are correct since a Java string is UTF-16, but this codec should operate on UTF-8.

xeno6696:
@noloader  I'm a bit scared of the --force push commits.  Typically with git it's a major red flag.  

What was necessitating the use of the command?

noloader:
Hi @xeno6696,

> I'm a bit scared of the --force push commits. Typically with git it's a major red flag
>
> What was necessitating the use of the command?

Rebasing to flatten out history to one commit.

I don't think it effects ESAPI's history. It is only effecting my fork.

kwwall:
@xeno6696 wrote:
> @kwwall It's not a make/break feature as in the past I've just had people do javascript escapes for JSON payload data. It's overkill but it works. This one is lighter weight.

Plus somewhat ironically, not EVERY use of JSON is associated with JavaScript any more. (At least in the usual sense of JavaScript use.) JSON is now one of the many ubiquitous information preserving file formats used much like CSV and XML files are to move information around, so I think people would favor something lighter weight.

noloader:
> Let's just keep it simple. I think that IllegalArgumentException works fine.

Perfect, thanks.

planetlevel:
Perhaps an Observable type model would be better for intrusion relevant events.

—Jeff

Jeff Williams
410-707-1487


________________________________

noloader:
> not EVERY use of JSON is associated with JavaScript any more...

++. Javascript encoding is not valid JSON encoding. A conforming JSON parser should reject `\x2f` and friends.

kwwall:
> Perhaps an Observable type model would be better for intrusion relevant events.

Hindsight is 20/20, but something to consider in ESAPI 3.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+     * @param immune character array of whitelist characters which should not be encoded
+     * @param c the character to encode if not in the immune list
+     * @return encoded character if the character is special, and the character otherwise.
+     */
+    public String encodeCharacter( char[] immune, int codePoint )
+        throws IllegalArgumentException {
+
+        // Per the docs for HTMLEntityCodec: "WARNING: This method will silently discard
+        // invalid code points per the call to Character.isValidCodePoint( int ) method.
+        // WARNING!! Character based Codecs will silently transform code points that are
```
kwwall, line 95: I think this may still be likely even if you stored them as int arrays because of the previously mentioned subtle truths about Java strings as UTF-16, but I'm not as up on this as you and @xeno6696 probably are. I try not to make my brain hurt by knowing more than I need to about code points. Instead, I prefer to concentrate on relatively simple stuff in comparison, like cryptography and quantum mechanics.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+        // not legal UTF code points into garbage data as they will cast them to chars.
+        if ( Character.isValidCodePoint( codePoint ) == false ) {
+            // throw new IllegalArgumentException( "Invalid codepoint '" + String.format("0x%04X", codePoint) + "'." );
```
kwwall, line 98: So, because we silently swallowed these before in other codecs, we are going to continue to do the same here? Wouldn't it make more sense to fix the other (IMO) misbehaved codecs and throw like you seemed to be planning to do here?

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+            return "";
+        }
+
+        if ( immune != null ) {
+            // More efficient than sort and binary search. If the immune array
+            // was presorted, then this could be O(log n). But we can't add the
+            // precondition now. It is too late in the game.
+            for ( Character ch : immune ) {
```
kwwall, line 106: I've never seen an immune array that was more than maybe 5 or 6 elements, so sorting would just be a waste of time unless you want to have a separate branch in that sorts if the # of immune elements would make the sorting worth while. Or instead you could just write:
```java
if ( immune.length > SOME_MAGIC_CONSTANT  ) {
    throw new ProgrammerHaveYouLostYourFreakinMindException("Too many immune characters; max is " + SOME_MAGIC_CONSTANT);
}
```

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+            case '\n': return "\\n";
+            case '\t': return "\\t";
+            case '"':  return "\\\"";
+            case '/':  return  "\\/";
+            case '\\': return "\\\\";
+        }
+
+        // Per the RFC... All Unicode characters may be placed within the
+        // quotation marks, except for the characters that MUST be escaped:
+        // quotation mark, reverse solidus, and the control characters
```
kwwall, line 128: Is there any really good reason to be obtuse and not just write '\' or even 'backslash' rather than 'reverse solidus' here? While documentation should be precise, it also should be understandable by the majority of practitioners.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+
+
+    /**
+     * {@inheritDoc}
+     *
+     * Decodes special characters in encoded JSON strings.
+     *
+     * decodeCharacter will decode the encoded character sequences for popular characters
+     * Backspace (\b), Form Feed (\f), Carriage Return (\r), Line Feed (\n), Tab (\t),
+     * Double Quote ("), Forward slash (/) and Backslash (\). The function will also decode
```
kwwall, line 146: You called it backslash here.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+                case 'n': return (int)'\n';
+                case 't': return (int)'\t';
+                case '"': return (int)'\"';
+                case '/': return  (int)'/';
+                case '\\': return (int)'\\';
+            }
+
+            errorMessage = "Invalid JSON two-character escape representation";
+
+            // Per the RFC... All characters may be escaped as a six-character sequence: a reverse solidus,
```
kwwall, line 192: Argh!

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+
+        // Catch all. The escaped character sequence was invalid.
+        input.reset();
+        throw new IllegalArgumentException( errorMessage );
+    }
+
+    protected int charToCodepoint( Character ch ) {
+
+        final String s = Character.toString(ch);
+        assert (s.length() == 1) : "Ooops";
```
kwwall, line 222: Since this is not a private method, this assertion (which are disabled by default) should really be a runtime check and throw an `IllegalArgumentException`...hopefully with a more meaningful message than 'Ooops'. ;-)

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -42,6 +42,7 @@
 import org.owasp.esapi.codecs.PercentCodec;
 import org.owasp.esapi.codecs.VBScriptCodec;
 import org.owasp.esapi.codecs.XMLEntityCodec;
+import org.owasp.esapi.codecs.JSONCodec;
```
kwwall, line 45: I don't personally care, but I almost will guarantee that if ESAPI is around for another dozen years, we'll get a PR that will alphabetize the import statements. I know, because I have seen that in some Java coding standards! :D

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1,15 +1,15 @@
 /**
  * OWASP Enterprise Security API (ESAPI)
- * 
+ *
  * This file is part of the Open Web Application Security Project (OWASP)
  * Enterprise Security API (ESAPI) project. For details, please see
  * <a href="http://www.owasp.org/index.php/ESAPI">http://www.owasp.org/index.php/ESAPI</a>.
```
kwwall, line 6: Fix the link and remove the pointless anchor tag since the standard doclet that generates Javadoc will not display this anyway so let's just make it easier for the future coders to read.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -38,11 +38,13 @@
 import org.owasp.esapi.codecs.HTMLEntityCodec;
 import org.owasp.esapi.codecs.MySQLCodec;
 import org.owasp.esapi.codecs.OracleCodec;
+import org.owasp.esapi.codecs.JSONCodec;
```
kwwall, line 41: Surprise, surprise. Someone already alphabetized these imports (or perhaps we can blame an IDE).

Seriously, your email tagline should be:
_I troll anal people._

kwwall:
@noloader - Just wondering, since you are doing this codec and JSON has become the linga franca of transporting data, for completeness, you might want to consider writing some new `Validator` methods, like `isValidSafeJSON()` that would be analogous to `isValidSafeHTML()`.

kwwall (commented):
I generally approve. Only real issue I saw was the use of assert 'Ooops' instead of a runtime check and throwing IllegalArgumentException. (I am guilty of the same thing and have changed many assertions in the last 5 years.)

I generally approve, but I'd like to take one more pass at resolving my comments before I merge this.

xeno6696:
> @noloader - Just wondering, since you are doing this codec and JSON has become the linga franca of transporting data, for completeness, you might want to consider writing some new `Validator` methods, like `isValidSafeJSON()` that would be analogous to `isValidSafeHTML()`.

AntiSamy does JSON now?

kwwall:
No, AFAIK, AntiSamy does NOT do JSON, which is why I asked if @noloader
wanted to do it.

xeno6696:
@kwwall  AntiSamy *only* handles HTML, that ask amounts to rewriting/cloning AntiSamy centering it around JSON parsing.  

And THEN you have the mixed context:  AntiSamy is designed to allow websites to safely handle user-inputted HTML.  Our use-case derives from that.  In what  context does a similar use case exist for JSON?  What scripts can you execute using JSON’s non-existent script tag?  What are you defending against, precisely?  What would be the json equivalent to the tag whitelisting AntiSamy does?

It’s better viewed as a transport layer data format, not a browser renderable scripting language like HTML or JavaScript. JSON could handle data intended to be used in  SQL or more likely, javascript itself or HTML intended to be rendered by innerHTML in javascript.  

The safeValidHTML contract works because HTML is the single explicit target.  With JSON you have everything.  You’re asking for an XSS servlet filter.  To drill that point home, why don’t we have a safeValidXML call which would be the closest equivalent to JSON?

planetlevel:
I generally agree with Matt here – since JSON and XML are just data they can’t contain attacks so AntiSamy wouldn’t be applicable.

Except that it’s not quite true.  XML can have doctypes that contain XXE attacks – would certainly be a nice thing to strip out.  And JSON is actually javascript and some people might eval it.  So you could imagine sanitizing it so that only data is left.

I’m not recommending that we build it, because it’s hard.  You probably couldn’t use an existing parser, because it might execute the code in the data.  So you’d have to build a custom one.  Which is quite difficult to get right.   I’d steer folks away from trying to sanitize XML and JSON and simply ensuring that those data formats are used safely.

--Jeff



From: Matt Seil ***@***.***>
Date: Wednesday, July 27, 2022 at 9:21 AM
To: ESAPI/esapi-java-legacy ***@***.***>
Cc: Jeff Williams ***@***.***>, Comment ***@***.***>
Subject: Re: [ESAPI/esapi-java-legacy] JSON string escaping support (PR #722)

@kwwall<https://github.com/kwwall> AntiSamy only handles HTML, that ask amounts to rewriting/cloning AntiSamy centering it around JSON parsing.

And THEN you have the mixed context: AntiSamy is designed to allow websites to safely handle user-inputted HTML. Our use-case derives from that. In what context does a similar use case exist for JSON? What scripts can you execute using JSON’s non-existent script tag? What are you defending against, precisely? What would be the json equivalent to the tag whitelisting AntiSamy does?

It’s better viewed as a transport layer data format, not a browser renderable scripting language like HTML or JavaScript. JSON could handle data intended to be used in SQL or more likely, javascript itself or HTML intended to be rendered by innerHTML in javascript.

The safeValidHTML contract works because HTML is the single explicit target. With JSON you have everything. You’re asking for an XSS servlet filter. To drill that point home, why don’t we have a safeValidXML call which would be the closest equivalent to JSON?

xeno6696:
@kwwall the safest way to accomplish that would be to use javacc to build a JSON parser of our own.  Nobody rolls their own parsers and lexers these days.  [This is incomplete but ](http://www.engr.mun.ca/~theo/JavaCC-Tutorial/javacc-tutorial.pdf) is where I usually begin when I sit down to do one of these.  (It's been 8yrs since the last time I did it.)  

@planetlevel has disliked the idea in the past because browsers for example are notorious for breaking BNF rules by trying to "help" by doing things like fixing invalid HTML.  That's valid.  But I think the risk would be less in the case of JSON because of its simplicity and lack of things like XXE attacks.  Also, we should call out that the best we can do is ensure that the JSON is legally parseable--which is what you're suggesting with your GSON example.

PR #736, opened 2022-08-21
Issue#735 ConfigurationException message changed to a more intuitive one

HenriquePinto333:
It was asked to change the message given by an exception in the EsapiPropertyLoaderFactory Class that specified that only .xml files or .properties files were allowed. 
So, using a user-friendly approach, I decided to call the ending of the file an "extension" and told in the message that only files with the  .xml or .properties "file extension" were supported.

### DIFF
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoaderFactory.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoaderFactory.java
@@ -52,7 +52,7 @@ public class EsapiPropertyLoaderFactory {
             return new StandardEsapiPropertyLoader(cfgPath, cfg.getPriority());
         } else {
             throw new ConfigurationException("The extension of given configuration path [ " + cfgPath + " ] is not supported." +
-                    "\nOnly .xml or .properties file extensions are supported.");
+                    "Only .xml or .properties file extensions are supported.");
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoaderFactory.java
```diff
@@ -51,8 +51,8 @@ public static AbstractPrioritizedPropertyLoader createPropertyLoader(EsapiConfig
         if (PROPERTIES.getTypeName().equalsIgnoreCase(fileExtension)) {
             return new StandardEsapiPropertyLoader(cfgPath, cfg.getPriority());
         } else {
-            throw new ConfigurationException("Configuration storage type [" + fileExtension + "] is not " +
-                    "supported");
+            throw new ConfigurationException("The extension of given configuration path [ " + cfgPath + " ] is not supported." +
+                    "\nOnly .xml or .properties file extensions are supported.");
```
kwwall, line 55: Minor change request. Either remove the '\n' here entirely or replace it with the value of
```java
System.getProperty("line.separator")
```
so that it works as intended one *nix and Windows. Personally, I'd just remove the newline and replace it with a space as that make it easier to grep logged exceptions.

kwwall (changes requested):
Otherwise, looks fine. Thanks for doing this.

src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoaderFactory.java
HenriquePinto333: Sure thing! Should be quick.

HenriquePinto333: I'm sorry for such a delay. I'm still new to git and github and this is my first pull request ever. I just finished my first year of college and all projects I have done were in a single branch and weren't forked, however I have already changed what you told me to change.

kwwall: @HenriquePinto333 - No worries. That's how we all learn. Took me way too long to get used to git after using at least a half-dozen other version control systems.

Issue #740, opened 2022-09-08
Update SLF4J log bridge to allow NULL EventTypes


Jeff-Walker:
All of the older log bridges allowed a `null` be passed as the first parameter of the log call. Because of this, we have *tons* of legacy code that pass a `null` when they really should have passed `Logger.EVENT_UNSPECIFIED`. The original log4j bridge allowed it and the newer JUL bridge does.

I will submit a PR for the log bridge to do a `null` check and choose a sane value based on the log level.

I have a custom log bridge and factory, but I had to cut and paste a lot of code from the official SLF4J code that I'd really rather have this in the code base if possible.

kwwall:
@Jeff-Walker - Thanks for making us aware of this and volunteering to do a PR. Could I ask you to please assign this ticket to yourself so that we know it's being worked on?

Lastly, FWIW, I am going to be submitting a PR in the next couple of days that updates ESAPI to use AntiSamy 1.7.1. And since that new AntiSamy version uses `slf4j:slf4j-api:2.0.0`, I will be updating our pom to use that as well. That probably won't make a difference, but if you start testing your PR before my soon-to-be-submitted PR is merged, you might want to make sure it works with `slf4j:slf4j-api:2.0.0`.

Jeff-Walker:
Here's the PR 
https://github.com/ESAPI/esapi-java-legacy/pull/741

Just let me know if I need to review this for the upgrade.

Jeff-Walker:
Looks like I don't have permission to assign myself.

kwwall:
Strange. I can't assign you either. NBD though since you have already
submitted the PR. I'll look at it tonight.

PR #741, opened 2022-09-08
issue-740

Jeff-Walker:
Adds null-safe EventType->Marker translation.
Adds unit test for new behavior.

fixes: https://github.com/ESAPI/esapi-java-legacy/issues/740

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogBridgeImpl.java
+++ b/src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogBridgeImpl.java
@@ -58,6 +58,7 @@ public class Slf4JLogBridgeImpl implements Slf4JLogBridge {
             throw new IllegalArgumentException("Unable to lookup SLF4J level mapping for esapi value of " + esapiLevel);
         }
         if (handler.isEnabled(logger)) {
+            type = safeEventType(type, esapiLevel);
             String fullMessage = appender.appendTo(logger.getName(), type, message);
             String cleanString = scrubber.cleanMessage(fullMessage);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogBridgeImpl.java
```diff
@@ -58,20 +58,46 @@ public void log(Logger logger, int esapiLevel, EventType type, String message) {
             throw new IllegalArgumentException("Unable to lookup SLF4J level mapping for esapi value of " + esapiLevel);
         }
         if (handler.isEnabled(logger)) {
+            type = safeEventType(type, esapiLevel);
```
jeremiahjstacey, line 61: My opinion is that this should account for null, and only null. .

`type = type == null? org.owasp.esapi.Logger.EVENT_UNSPECIFIED : type;`

The additional types for ERROR and Fatal are  already being accounted for in the delegate Handler being called, and will be reflected in the level at which the SLF4J logger records the event.

I don't think the additional evaluation of the esapiLevel value is needed.

(borrowed from the SLF4JFactory class)
```
 Map<Integer, Slf4JLogLevelHandler> levelLookup = new HashMap<>();
        levelLookup.put(Logger.ALL, Slf4JLogLevelHandlers.TRACE);
        levelLookup.put(Logger.TRACE, Slf4JLogLevelHandlers.TRACE);
        levelLookup.put(Logger.DEBUG, Slf4JLogLevelHandlers.DEBUG);
        levelLookup.put(Logger.INFO, Slf4JLogLevelHandlers.INFO);
        levelLookup.put(Logger.ERROR, Slf4JLogLevelHandlers.ERROR);
        levelLookup.put(Logger.WARNING, Slf4JLogLevelHandlers.WARN);
        levelLookup.put(Logger.FATAL, Slf4JLogLevelHandlers.ERROR);
```

jeremiahjstacey: Additionally, I do not think that null is handled in the JavaLogBridgeImpl either.  For consistency, I think it would be good to apply the ternary check in that class as well to prevent the same problem from being repeated in that scope.

Jeff-Walker: Comment 1: that's fair. I'll remove and resubmit.
Comment 2: I'm pretty sure it does handle null, because my old code works. It's because the bridge impl doesn't do anything with the event type object but pass it to the appender, who handles the null no problem. The only reason that the SLF4J bridge has a problem is that it calls toString to make the marker object. I will go over it to make sure.

Jeff-Walker:
I made the update. I also added a test to the JUL bridge to make sure it doesn't throw an exception on a null event type.

Jeff-Walker:
Cool, thanks. 
Do you have a target date for the next release yet?

kwwall:
@Jeff-Walker - TBH there hasn't been a lot done since the 2.5.0.0 release on 7/20/2022 and there is a significant time sink in doing the release (typically 2 hrs minimum plus whatever it takes to write up the release notes). Admittedly, the release notes would be comparatively simple this time. I would ideally like to knock out a few more bug fixes before the next release, so I think something around end of Oct / beginning of Nov would be more realistic.

Jeff-Walker:
Thanks again.

PR #744, opened 2022-10-01
Issue 710 JUL Respect LogManager Config

### LINKED ISSUE #710: JavaLogFactory configuration should not override custom java LogManager configuration
### Discussed in https://github.com/ESAPI/esapi-java-legacy/discussions/709

<div type='discussions-op-text'>

<sup>Originally posted by **chuckdumont** June 17, 2022</sup>
Our application uses java.util.logging and handles logging configuration (setting handlers, log levels, etc.).  When using esapi with JUL logging, the application's configured log levels are reset by esap.  How can we configure esapi to use the java.util.logging APIs to log events, but not try to set it's own logging config?</div>

#### Recommended Adjustment
If either supported configuration property for the LogManager class exists in the runtime, ESAPI will not apply configuration settings from the esapi-java-logging.properties file

Per the javadoc on the [LogManager](https://docs.oracle.com/javase/7/docs/api/java/util/logging/LogManager.html) class, two existing properties may be used to centralize the log configuration:
> In addition, the LogManager uses two optional system properties that allow more control over reading the initial configuration:
>
>  - "java.util.logging.config.class"
>  - "java.util.logging.config.file"
>
> These two properties may be set via the Preferences API, or as command line property definitions to the "java" command, or as system property definitions passed to JNI_CreateJavaVM.

The ESAPI JavaLogFactory should not conflict with those settings, if they are defined.  

#### JavaLogFactory.readLoggerConfiguration Additions
```
  /*package*/ static void readLoggerConfiguration(LogManager logManager) {
        if (System.getProperties().keySet().stream().anyMatch(propKey -> "java.util.logging.config.class".equals(propKey) || "java.util.logging.config.file".equals(propKey))) {
            //LogManager has external configuration.  Do not load ESAPI defaults.
            return;
        }
```

jeremiahjstacey:
Closes #710 

Updating JUL to check for existing global LogManager configuration prior to applying ESAPI file settings.

Adding tests to verify LogManager Configurations are preferred to the ESAPI configurations.

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -48,9 +48,9 @@ import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
  * Options for customizing this configuration (in recommended order)
  * <ol>
  * <li>Consider using the <i>SLF4JLogFactory</i> with a java-logging implementation.</li>
- * <li>Configure java LogManager system properties as defined by the <i>java.util.logging.LogManager</i> API</li>
+ * <li>Configure the runtime startup command to set the desired system properties for the <i>java.util.logging.LogManager</i> instance.  EG: <code>-Djava.util.logging.config.file=/custom/file/path.properties</code></li>
  * <li>Overwrite the esapi-java-logging.properties file with the desired logging configurations. <br>A default file implementation is available in the configuration jar on GitHub under the 'Releases'</li>
- * <li>Specifically set the system property for LogManager in the custom-code solution to prevent ESAPI JavaLogFactory from overriding project configuration</li>
+ * <li>Apply custom-code solution to set the system properties for the <i>java.util.logging.LogManager</i> at runtime. EG: <code>System.setProperty("java.util.logging.config.file", "/custom/file/path.properties");</code></li>
  * <li>Create a custom JavaLogFactory class in client project baseline and update the ESAPI.properties configuration to use that reference.</li>
  * </ol>
  *
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -47,7 +47,7 @@ import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
  * Options for customizing this configuration (in recommended order)
  * <ol>
  * <li>Consider using the <i>SLF4JLogFactory</i> with a java-logging implementation.</li>
- * <li>Configure java LogManager environment properties as defined by the <i>java.util.logging.LogManager</i> API</li>
+ * <li>Configure java LogManager system properties as defined by the <i>java.util.logging.LogManager</i> API</li>
  * <li>Overwrite the esapi-java-logging.properties file with the desired logging configurations. <br>A default file implementation is available in the configuration jar on GitHub under the 'Releases'</li>
  * <li>Specifically set the system property for LogManager in the custom-code solution to prevent ESAPI JavaLogFactory from overriding project configuration</li>
  * <li>Create a custom JavaLogFactory class in client project baseline and update the ESAPI.properties configuration to use that reference.</li>
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -25,6 +25,7 @@ import java.util.logging.LogManager;
 import org.owasp.esapi.ESAPI;
 import org.owasp.esapi.LogFactory;
 import org.owasp.esapi.Logger;
+import org.owasp.esapi.PropNames;
 import org.owasp.esapi.codecs.HTMLEntityCodec;
 import org.owasp.esapi.errors.ConfigurationException;
 import org.owasp.esapi.logging.appender.LogAppender;
@@ -102,6 +103,18 @@ public class JavaLogFactory implements LogFactory {
         "java.util.logging.config.class".equals(propKey) || "java.util.logging.config.file".equals(propKey))) {
             // LogManager has external configuration.  Do not load ESAPI defaults.
             // See javadoc for the LogManager class for more information on properties.
+            boolean isStartupSysoutDisabled = Boolean.valueOf(System.getProperty(PropNames.DISCARD_LOGSPECIAL, Boolean.FALSE.toString()));
+            if (!isStartupSysoutDisabled) {
+                String logManagerPreferredMsg = String.format("[ESAPI-STARTUP] ESAPI JavaLogFactory Configuration will not be applied. "
+                        + "java.util.LogManager configuration Detected. "
+                        + "{\"java.util.logging.config.class\":\"%s\",\"java.util.logging.config.file\":\"%s\"}",
+                        System.getProperty("java.util.logging.config.class"), System.getProperty("java.util.logging.config.file"));
+
+                System.out.println(logManagerPreferredMsg);
+                // ::SAMPLE OUTPUT::
+                //[ESAPI-STARTUP] ESAPI JavaLogFactory Configuration will not be applied.  java.util.LogManager configuration Detected.{"java.util.logging.config.class":"some.defined.value","java.util.logging.config.file":"null"}
+            }
+
             return;
         }
         /*
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -43,10 +43,15 @@ import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
 
 /**
  * LogFactory implementation which creates JAVA supporting Loggers.
- *
- * This implementation requires that a file named 'esapi-java-logging.properties' exists on the classpath.
- * <br>
- * A default file implementation is available in the configuration jar on GitHub under the 'Releases'
+ * <br><br>
+ * Options for customizing this configuration (in recommended order)
+ * <ol>
+ * <li>Consider using the <i>SLF4JLogFactory</i> with a java-logging implementation.</li>
+ * <li>Configure java LogManager environment properties as defined by the <i>java.util.logging.LogManager</i> API</li>
+ * <li>Overwrite the esapi-java-logging.properties file with the desired logging configurations. <br>A default file implementation is available in the configuration jar on GitHub under the 'Releases'</li>
+ * <li>Specifically set the system property for LogManager in the custom-code solution to prevent ESAPI JavaLogFactory from overriding project configuration</li>
+ * <li>Create a custom JavaLogFactory class in client project baseline and update the ESAPI.properties configuration to use that reference.</li>
+ * </ol>
  *
  */
 public class JavaLogFactory implements LogFactory {
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -53,6 +53,8 @@ import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
  * <li>Apply custom-code solution to set the system properties for the <i>java.util.logging.LogManager</i> at runtime. EG: <code>System.setProperty("java.util.logging.config.file", "/custom/file/path.properties");</code></li>
  * <li>Create a custom JavaLogFactory class in client project baseline and update the ESAPI.properties configuration to use that reference.</li>
  * </ol>
+ * 
+ * @see <a href="https://github.com/ESAPI/esapi-java-legacy/wiki/Configuration-Reference:-JavaLogFactory">ESAPI Wiki - Configuration Reference: JavaLogFactory</a>
  *
  */
 public class JavaLogFactory implements LogFactory {
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -93,6 +93,12 @@ public class JavaLogFactory implements LogFactory {
      * @param logManager LogManager which is being configured.
      */
     /*package*/ static void readLoggerConfiguration(LogManager logManager) {
+        if (System.getProperties().keySet().stream().anyMatch(propKey ->
+        "java.util.logging.config.class".equals(propKey) || "java.util.logging.config.file".equals(propKey))) {
+            // LogManager has external configuration.  Do not load ESAPI defaults.
+            // See javadoc for the LogManager class for more information on properties.
+            return;
+        }
         /*
          * This will load the logging properties file to control the format of the output for Java logs.
          */
--- a/src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
+++ b/src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
@@ -49,6 +49,59 @@ public class JavaLogFactoryTest {
     public ExpectedException exEx = ExpectedException.none();
 
     @Test
+    public void testLogManagerConfigurationAsClass() throws Exception {
+        String propKey = "java.util.logging.config.class";
+        //If defined, grab the value; otherwise, set to a known value to allow for prop to be cleared.
+        String sysDefault = System.getProperties().stringPropertyNames().contains(propKey) ? System.getProperty(propKey) : testName.getMethodName();
+
+        System.setProperty(propKey, "some.defined.value");
+        LogManager testLogManager = new LogManager() {
+            @Override
+            public void readConfiguration(InputStream ins) throws IOException, SecurityException {
+                throw new IOException(testName.getMethodName());
+            }
+        };
+
+        try {
+            // This would throw an IOException if the LogManager was not being respected since no esapi-java-logging file is specified
+            JavaLogFactory.readLoggerConfiguration(testLogManager);
+        } finally {
+            //Restore original prop values
+            if (testName.getMethodName().equals(sysDefault))
+                System.clearProperty(propKey);
+            else {
+                System.setProperty(propKey, sysDefault);
+            }
+        }
+    }
+
+    @Test
+    public void testLogManagerConfigurationAsFile() throws Exception {
+        String propKey = "java.util.logging.config.file";
+        //If defined, grab the value; otherwise, set to a known value to allow for prop to be cleared.
+        String sysDefault = System.getProperties().stringPropertyNames().contains(propKey) ? System.getProperty(propKey) : testName.getMethodName();
+
+        System.setProperty(propKey, "some.defined.value");
+        LogManager testLogManager = new LogManager() {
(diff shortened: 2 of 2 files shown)


src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -43,10 +43,15 @@
...
  * LogFactory implementation which creates JAVA supporting Loggers.
- *
- * This implementation requires that a file named 'esapi-java-logging.properties' exists on the classpath.
- * <br>
- * A default file implementation is available in the configuration jar on GitHub under the 'Releases'
+ * <br><br>
+ * Options for customizing this configuration (in recommended order)
+ * <ol>
+ * <li>Consider using the <i>SLF4JLogFactory</i> with a java-logging implementation.</li>
+ * <li>Configure java LogManager environment properties as defined by the <i>java.util.logging.LogManager</i> API</li>
```
kwwall, line 50: I don't see any mention of "environment variables" (which generally Java tries to avoid) in the Javadoc for [java.util.logging.LogManager](https://docs.oracle.com/javase/8/docs/api/java/util/logging/LogManager.html). I only say the 2 system properties which you see to account (which are not specifically mentioned by name here, but probably should be in case others are added in the future) for in the list item below.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -93,6 +98,12 @@ public class JavaLogFactory implements LogFactory {
      * @param logManager LogManager which is being configured.
      */
     /*package*/ static void readLoggerConfiguration(LogManager logManager) {
+        if (System.getProperties().keySet().stream().anyMatch(propKey ->
```
kwwall, line 101: So here's my concern and I'm not sure of the best way to address it. If a developer configures JUL's LogManager themselves, but show how mess it up because say, they botch the name of their config file or it can't be found, etc., we are going to end up getting emails,GitHub issues, GitHub Discussions, and/or Stack Overflow. How can we prevent this or at least reduce this possibility? At the very least, I think we should have some notice in a comment around the **ESAPI.Logger** property in **configuration/esapi/ESAPI.properties**. But if they do botch one of these 2 JUL system properties, we should probably test to see what that exception stack looks like and perhaps document it somewhere, perhaps somewhere under our local GitHub wiki.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
```diff
@@ -48,6 +48,59 @@ public class JavaLogFactoryTest {
...
+                System.clearProperty(propKey);
+            else {
+                System.setProperty(propKey, sysDefault);
+            }
+        }
+    }
+
+    @Test
+    public void testLogManagerConfigurationAsFile() throws Exception {
+        String propKey = "java.util.logging.config.file";
```
kwwall, line 80: What if this property is set, but the file can't be found? I suspect that it might have consequences down the line when someone actually attempts to use the ESAPI JUL logger. Is that something that can test and see how it affects us and at least document it somewhere?

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: from the LogManager API linked:
```
The LogManager defines two optional system properties that allow control over the initial configuration:

"java.util.logging.config.class"
"java.util.logging.config.file"
```

I will update line 50 to refer to them as "system properties" to match the language in the class' javadoc

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: There are so many ways to misconfigure a logger, I'd prefer to not do this if we can avoid it.

- File presence
- file permissions
- file contents (xml, properties configurations)
- Class reference availability (classpath configuration)
- Class reference readability (security Manager configuration)

Just to name a few.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
jeremiahjstacey: IMO that is a test that the Java library is responsible for.  Once we've determined that ESAPI will not be loading the relative file for logging configurations we've fulfilled the responsibility of what we can influence.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
kwwall: I was thinking perhaps of putting it in a try-block and then doing something along the lines of:
```java 
} catch(Exception ex) {
    StackTraceElement ste[] = ex.getStackTrace();
    // enumerate through the stack trace to see if the exception was caused by ESAPI or not
}
```
That would take a lot of experimentation though. Maybe better is just to use logSpecial() just before the return at line 100 and note (once only!) something to state that "Any screw-ups regarding logging are not the fault of ESAPI". E.g.,
```java
if ( warningCounter == 0 ) {
    logSpecial("WARNING: Here be dragons! You have chosen to take full control of JUL configuration used with ESAPI logging. Any failure of ESAPI to log correctly is on you, the developer, and not the fault of ESAPI so please don't post questions or issues about it.");
    warningCounter++'
}
```
or something to that effect. Not really sure what the contents are.

jeremiahjstacey: The try/catch would not catch anything at this point, which is part of the difficulty. My understanding is that the LogManager will load this once on initialization.  We're not delegating to the LogManager or relying on any calls to it here.  Additionally, even if we were to add something to that effect it may not even trigger if ESAPI isn't the first reference to attempt to initialize the LogManager.

jeremiahjstacey: adding a System.out.println / logSpecial would be good.  It at least leaves a breadcrumb of the workflow that was applied to the environment.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
kwwall: My point was we're pretty sure that it will cause an exception of some sort, but if the first attempt at JUL logging is via ESAPI, the devs will likely blame ESAPI because our classes will be sitting at the top of the stack. Lower down, there probably will be a ''caused by:" that they will overlook. So it would be good for us to test it (even if we don't leave that test in) and document it somewhere so that when we inevitably get these questions / issues, we don't have to research the problem.

I think more likely what will cause a problem is if the don't configure the properties that you need that our Java logger needs to work properly that normally would be picked up from the esapi-java-logging.properties configuration file in our configuration jar. If all of these properties:
```
handlers
.level
java.util.logging.ConsoleHandler.level
java.util.logging.ConsoleHandler.formatter
java.util.logging.SimpleFormatter.format
```
are mandatory with JUL, then we shouldn't have a problem. But if any of they are optional, we definitely ought to note that in the Javadoc and perhaps in the ESAPI.properties and anywhere else that people might see it.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -43,10 +43,15 @@
...
- * This implementation requires that a file named 'esapi-java-logging.properties' exists on the classpath.
- * <br>
- * A default file implementation is available in the configuration jar on GitHub under the 'Releases'
+ * <br><br>
+ * Options for customizing this configuration (in recommended order)
+ * <ol>
+ * <li>Consider using the <i>SLF4JLogFactory</i> with a java-logging implementation.</li>
+ * <li>Configure java LogManager environment properties as defined by the <i>java.util.logging.LogManager</i> API</li>
+ * <li>Overwrite the esapi-java-logging.properties file with the desired logging configurations. <br>A default file implementation is available in the configuration jar on GitHub under the 'Releases'</li>
+ * <li>Specifically set the system property for LogManager in the custom-code solution to prevent ESAPI JavaLogFactory from overriding project configuration</li>
```
kwwall, line 52: I'm not clear what you mean by this. More specifically, what is different about this than about the comment at line 50 (other than those are system properties and not environment variables).

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
kwwall: I like this idea (even though people can and probably do, disable the logSpecial() output. Let's just not make sure something is printed a gazillion times. Once is enough. If they ignore it, that's on them.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: This change has been applied.   Still looking at addressing the other content around error handling and visibility of LogManager configuration.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: logSpecial content has been added to this block Current message format is below:
`[ESAPI-STARTUP] ESAPI JavaLogFactory Configuration will not be applied.  java.util.LogManager configuration Detected.{"java.util.logging.config.class":"null","java.util.logging.config.file":"some.defined.value"}`

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: @kwwall, I think you're referring to line 52 vs 50?
>[50] Configure java LogManager environment properties as defined by the <i>java.util.logging.LogManager</i> API

and

> [52] Specifically set the system property for LogManager in the custom-code solution to prevent ESAPI JavaLogFactory from overriding project configuration


My intent is that line 50 would be adding the appropriate `-D` to the application initialization.
Line 52 is more referring to adding the call to `System.setProperty` in the application client code.

The end state should be about the same, as long as the `System.setProperty` invocation happens before ESAPI starts up.  That race condition is why it is lower on the recommended order.  It *might* work, but someone needs to understand the application initialization order.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
jeremiahjstacey: Still poking at this.   I'm not getting the output I was expecting from testing.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -48,9 +48,9 @@
  * Options for customizing this configuration (in recommended order)
  * <ol>
  * <li>Consider using the <i>SLF4JLogFactory</i> with a java-logging implementation.</li>
- * <li>Configure java LogManager system properties as defined by the <i>java.util.logging.LogManager</i> API</li>
+ * <li>Configure the runtime startup command to set the desired system properties for the <i>java.util.logging.LogManager</i> instance.  EG: <code>-Djava.util.logging.config.file=/custom/file/path.properties</code></li>
```
jeremiahjstacey, line 51: @kwwall  Do these updates make the intended difference more clear?

kwwall:
Yes, I think so.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
jeremiahjstacey: This is not forgotten.   I have not had time to look more into it yet.  Hoping to have a chance this weekend.

jeremiahjstacey: @kwwall specifically to your comment above on the content of esapi-java-logging.properties, the file contains the minimum content to get log output.  All defined content is required for basic application output.

I have tested some variations of misconfiguration for LogManger, and I'm just not seeing the results I would expect.  That's making it difficult for me to offer any technical scale of FAQ that would be useful for this thread.  Instead, I've collected my current thoughts and understandings as a 'primer' to assist users getting started if they experience trouble.

I'm not sure where to take this from here.

-----

## Configure ESAPI to use the JavaLogFactory
The ESAPI.properties file must contain the required key `ESAPI.Logger`, and the value must be configured to the fully-qualified name of the `org.owasp.esapi.LogFactory` implementation to be used.

In the most recent releases of the ESAPI library the JavaLogFactory is the default logging configuration set by the project:
`ESAPI.Logger=org.owasp.esapi.logging.java.JavaLogFactory`

## Configure the Java Logging Output
ESAPI supports two primary methods for configuring the output from the Java Log implementation.  The application must supply at least one of these configurations.  Upon initialization, the ESAPI JavaLogFactory will make a best-effort attempt to verify that the Java Log behavior has some configuration in place.

### Application-Wide LogManger Configuration
If either of the following system properties are detected at runtime ESAPI will not look for, or load, the esapi-java-logging.properties file.
`java.util.logging.config.class`
`java.util.logging.config.file`

Note these system properties are a part of Java and not owned, controlled, or managed by ESAPI.
[JDK8 LogMaganger Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/logging/LogManager.html)


Since ESAPI must be able to detect the configurations, they must be present as runtime configurations.

**Startup System Properties**
Most java application provide a way to extend the startup configurations/options for the application instance. The exact process and configuration location is dependent on what application is being used.  Refer to the application administration guide or documentation as to how to configure this correctly for the targeted environment. 

`-Djava.util.logging.config.class= `
`-Djava.util.logging.config.file= `

ESAPI takes no ownership or responsibility for diagnosing application-specific configuration requirements.

**In-Code Solutions** (Not Recommended)
Java provides a mechanism to set System properties in code at runtime.  By calling the `System.setProperty` method, the desired configuration can be applied over the course of normal application exection.

`System.setProperty('java.util.logging.config.class", "");`
`System.setProperty('java.util.logging.config.file", "");`

**NOTE** this method has a heavy dependency on application startup order.  The calls to set the system properties must be guaranteed to run before the first call to ESAPI's log is made.  Due to this unpredicatable state this method is not recommended for complex applications or any system with dynamic startup execution order.

### ESAPI Configuration
1. Download the  esapi-{version}-configuration.jar from the [ESAPI Releases](https://github.com/ESAPI/esapi-java-legacy/releases) page 
2. Extract the default esapi-java-logging.properties file from the configuration jar.
3. Place the properties file next to `ESAPI.properties` on your application's path.
4. Restart the application.

## Resources

### Debugging Failed Startup
In the case where the JavaLogFactory is unable to verify the java.logging.LogManger system properties and no esapi-java-logging.properties file exists the library will not load.

To verify, the application startup logs should be inspected for the content identified below.

```
Caused by: org.owasp.esapi.errors.ConfigurationException: Unable to locate resource: esapi-java-logging.properties
        at org.owasp.esapi.logging.java.JavaLogFactory.readLoggerConfiguration(JavaLogFactory.java:126)
        at org.owasp.esapi.logging.java.JavaLogFactory.<clinit>(JavaLogFactory.java:94)
        ... 47 more
```

**This content will only appear the first time the JavaLogFactory is requested.**

The JavaLogFactory class will not be loaded to the runtime, and will cause downstream exceptions containing `Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.owasp.esapi.logging.java.JavaLogFactory`

### Diagnosing Common Problems
#### I've Added esapi-java-logging.properties, and now other system log formats have changed!

If there is no detectable LogMaganger configuration, ESAPI reads a supplied esapi-java-logging.properties configuration and applies those configurations to the java LogManager for the runtime.  This can lead to race conditions on startup among projects that supply independent logging configurations, each one setting the local configuration as the java LogManager default.

To resolve this for ESAPI:  Set the LogMaganger system properties as identified above in **Configure the Java Logging Output/Application-Wide LogManger Configuration**

####  My application starts up, but I don't see any output in my logs!

First, verify that ESAPI did correctly start by looking for `Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.owasp.esapi.logging.java.JavaLogFactory` in your application logs.  See **Debugging Failed Startup** above.

If the application is starting up correctly and the logs are not working as desired, the the likely issue is in the configuration files.
  - If the application is using the esapi-java-logging.properties file, verify the file has not been modified.  Download the  esapi-{version}-configuration.jar from the ESAPI Releases page (linked below) and compare for any local modification.
  - If the application is using a more custom logging solution with an application-wide logging configuration, then verify that configuration.  The **Links** section below provides a reference to Oracle's Java Logging Overview that can be referenced for assistance.
    - If the In-Code Solution is being used, stop. Logging issues are likely a result of a dyamic application startup order.


### Links
 - [ESAPI Releases](https://github.com/ESAPI/esapi-java-legacy/releases)
   - this site can be used to download the esapi-{version}-configuration.jar for the targeted release.  It will contain a default implementation of ESAPI.properties and esapi-java-logging.properties
 - [Java Logging Overview](https://docs.oracle.com/javase/10/core/java-logging-overview.htm#JSCOR-GUID-B83B652C-17EA-48D9-93D2-563AE1FF8EDA)
   - Reference for customizing the Java Log setup and verifying required configuration

kwwall: @jeremiahjstacey - Wow. This looks great. Definitely along the lines that I was thinking. The only change to this documentation is that if we are going to do this, I'd like it placed on our GitHub wiki pages somewhere and referenced in the Javadoc and then take this statement:
> ESAPI takes no ownership or responsibility for diagnosing application-specific configuration requirements.

and move it up right after under the heading "Application-Wide LogManger Configuration" and make it more prominent (where are <blink> tags when you really need them?) by maybe writing it something like this:

### Application-Wide LogManger Configuration
_**WARNING**_: ESAPI takes no ownership or responsibility for diagnosing application-specific configuration requirements. This means that we may decline from answering questions about it in GitHub issues, our GitHub Discussions board, either of our ESAPI mailing lists, Stack Overflow, etc. and if we do answer questions about it, it likely will not go beyond referring to this wiki page and noting that "you have been warned".
blah, blah...

Well, you get the idea. (Tone it down or crank it up to your likely.)
Otherwise looks great.

kwwall:
@jeremiahjstacey  - Dang, I prematurely did 'Resolve conversation' without getting your feedback. Sorry about that. I guess you we can merge this PR and your can add a link to the Javadoc to the new GitHub wiki page in a separate PR. Or, if you want to add a new commit to that for this PR, let me know. I'll hold off merging this until I hear from you how you wish to handle it.

jeremiahjstacey:
[esapi_java_log_config.md](https://github.com/ESAPI/esapi-java-legacy/files/9827165/esapi_java_log_config.md)

Thanks for the feedback.  I will need to refresh my memory on how to interact with the wiki again -- been a little while for that.   I've integrated your suggestion into the content and attached the result (above) for safe-keeping until the wiki integration is complete.

jeremiahjstacey:
https://github.com/ESAPI/esapi-java-legacy/wiki/Configuration-Reference:-JavaLogFactory

Thanks!  That worked great. 

Javadoc is updated and pushed, but I can certainly adjust the page name if desired.

kwwall:
@jeremiahjstacey - The name of the wiki page is fine by me. Merging.

PR #753, opened 2022-10-26
JSON string escaping support

### LINKED ISSUE #754: JSON encoder
Implement a JSON output encoder to encode data for use in JSON strings as per [Section 7 of RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259#section-7).

**Is your feature request related to a problem? Please describe.**
@noloader needs it for his $dayjob.

**Describe the solution you'd like**
Provided as PR #753 

**Describe alternatives you've considered including other security libraries**
Using JavaScript to to the dirty work. That path was rejected because different implementations support slightly different grammars.

noloader:
This commit adds JSON string escaping support according to RFC 8259, Section 7.
See GitHub Issue 
closes #754. [added by @kwwall]

### DIFF
--- a/src/main/java/org/owasp/esapi/codecs/AbstractCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/AbstractCodec.java
@@ -58,8 +58,10 @@ public abstract class AbstractCodec<T> implements Codec<T> {
     /**
      * {@inheritDoc}
      * </p><p>
-     * <b>WARNING!!</b>  {@code Character} based {@code Codec}s will silently transform code points that are not
-     * legal UTF code points into garbage data as they will cast them to {@code char}s.
+     * <b>WARNING!!</b>  {@code Character} based {@code Codec}s will only handle the byte range of
+     * 0-65535 (0x0-0xffff).  Passing any data represented by a higher numerical value will result in
+     * a downcast thus destroying the original data with undefined results.
+     * <p>
      * Also, if you are implementing an {@code Integer} based codec, these will be silently discarded
      * based on the return from {@code Character.isValidCodePoint( int )}.  This is the preferred
      * behavior moving forward.
--- a/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
@@ -92,8 +92,10 @@ public class JSONCodec extends AbstractIntegerCodec {
 
         // Per the docs for HTMLEntityCodec: "WARNING: This method will silently discard
         // invalid code points per the call to Character.isValidCodePoint( int ) method.
-        // WARNING!! Character based Codecs will silently transform code points that are
-        // not legal UTF code points into garbage data as they will cast them to chars.
+
+        // WARNING!! Character based Codecs will only handle the byte range of 0-65535 (0x0-0xffff).
+        // Passing any data represented by a higher numerical value will result in a downcast thus
+        // destroying the original data with undefined results.
         if ( Character.isValidCodePoint( codePoint ) == false ) {
             // throw new IllegalArgumentException( "Invalid codepoint '" + String.format("0x%04X", codePoint) + "'." );
             return "";
--- a/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
@@ -155,65 +155,66 @@ public class JSONCodec extends AbstractIntegerCodec {
      *     if an invalid character sequence is encountered
      */
     public Integer decodeCharacter( PushbackSequence<Integer> input )
-        throws IllegalArgumentException {
+            throws IllegalArgumentException {
 
         input.mark();
 
         Integer first = input.next(), second = null;
+
+        //Ensure both first and second characters are not null before attempting to decode.
         if ( first == null || first.intValue() != '\\' ) {
             input.reset();
             return null;
         }
+        if ( (second = input.next()) == null ) {
+            throw new IllegalArgumentException( "Invalid JSON escape representation");
+        }
 
-        String errorMessage = null;
+        Integer decodedRef = null;
 
-        try
-        {
-            errorMessage = "Invalid JSON escape representation";
-
-            if ( (second = input.next()) == null ) {
-                throw new IllegalArgumentException();
-            }
-
-            // Per the RFC... Two-character sequence escape representations of some popular characters
-            switch ( second.intValue() ) {
-                case 'b': return (int)'\b';
-                case 'f': return (int)'\f';
-                case 'r': return (int)'\r';
-                case 'n': return (int)'\n';
-                case 't': return (int)'\t';
-                case '"': return (int)'\"';
-                case '/': return  (int)'/';
-                case '\\': return (int)'\\';
-            }
-
-            errorMessage = "Invalid JSON two-character escape representation";
-
+        // Per the RFC... Two-character sequence escape representations of some popular characters
+        switch ( second.intValue() ) {
+        case 'b':
+            decodedRef = (int)'\b';
+            break;
+        case 'f':
+            decodedRef = (int)'\f';
+            break;
+        case 'r':
+            decodedRef = (int)'\r';
+            break;
+        case 'n':
+            decodedRef = (int)'\n';
+            break;
+        case 't':
+            decodedRef = (int)'\t';
+            break;
+        case '"':
+            decodedRef = (int)'\"';
+            break;
+        case '/':
+            decodedRef =  (int)'/';
+            break;
+        case '\\':
+            decodedRef = (int)'\\';
+            break;
+        case 'u':
             // Per the RFC... All characters may be escaped as a six-character sequence: a reverse solidus,
             // followed by the lowercase letter u, followed by four hexadecimal digits that encode the
             // character's code point. The hexadecimal letters A through F can be uppercase or lowercase.
             // So, for example, a string containing only a single reverse solidus character may be represented
             // as "\u005C".
-            if ( second.intValue() == 'u' ) {
-
-                errorMessage = "Invalid JSON six-character escape representation";
-
-                return (convertToInt( input.next() ) << 12) +
-                       (convertToInt( input.next() ) <<  8) +
-                       (convertToInt( input.next() ) <<  4) +
-                       (convertToInt( input.next() ) <<  0);
-            }
-
-            // Do nothing. Fall into throw below.
-        }
-        catch (IllegalArgumentException e)
-        {
-            // Do nothing. Fall into throw below.
+            decodedRef = (convertToInt( input.next() ) << 12) +
+                (convertToInt( input.next() ) <<  8) +
+                (convertToInt( input.next() ) <<  4) +
+                (convertToInt( input.next() ) <<  0);
+            break;
+        default:
+            input.reset();
+            throw new IllegalArgumentException( "Invalid JSON two-character escape representation: " + String.format("'%s%s'", (char) first.intValue(), (char) second.intValue()) );
         }
 
-        // Catch all. The escaped character sequence was invalid.
-        input.reset();
-        throw new IllegalArgumentException( errorMessage );
+        return decodedRef;
     }
 
     protected int charToCodepoint( Character ch ) {
--- a/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
@@ -211,7 +211,7 @@ public class JSONCodec extends AbstractIntegerCodec {
             break;
         default:
             input.reset();
-            throw new IllegalArgumentException( "Invalid JSON two-character escape representation: " + String.format("'%s%s'", (char) first.intValue(), (char) second.intValue()) );
+            throw new IllegalArgumentException( "Invalid JSON two-character escape representation: " + String.format("'%c%c'", (char) first.intValue(), (char) second.intValue()) );
         }
 
         return decodedRef;
--- a/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
+++ b/src/main/java/org/owasp/esapi/codecs/JSONCodec.java
@@ -1,0 +1,242 @@
+/**
+ * OWASP Enterprise Security API (ESAPI)
+ *
+ * This file is part of the Open Web Application Security Project (OWASP)
+ * Enterprise Security API (ESAPI) project. For details, please see
+ * <a href="http://www.owasp.org/index.php/ESAPI">http://www.owasp.org/index.php/ESAPI</a>.
+ *
+ * Copyright (c) 2022 - The OWASP Foundation
+ *
+ * The ESAPI is published by OWASP under the BSD license. You should read and accept the
+ * LICENSE before you use, modify, and/or redistribute this software.
+ *
+ * @author Jeffrey Walton (noloader .at. gmail.com)
+ * @author Kevin Wall (kevin.w.wall .at. gmail.com)
+ * @author Matt Seil (matt.seil .at. owasp.org)
+ * @created 2022
+ */
+package org.owasp.esapi.codecs;
+
+/**
+ * Implementation of the Codec interface for JSON strings.
+ * This class performs <a
+ * href="https://datatracker.ietf.org/doc/html/rfc8259#section-7">String escaping</a>
+ * on the entire string according to RFC 8259, Section 7.
+ *
+ * RFC 8259 requires conforming implementations use UTF-8. However, the ESAPI interfaces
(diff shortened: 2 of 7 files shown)


src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+     */
+    public String encodeCharacter( char[] immune, int codePoint )
+        throws IllegalArgumentException {
+
+        // Per the docs for HTMLEntityCodec: "WARNING: This method will silently discard
+        // invalid code points per the call to Character.isValidCodePoint( int ) method.
+        // WARNING!! Character based Codecs will silently transform code points that are
+        // not legal UTF code points into garbage data as they will cast them to chars.
+        if ( Character.isValidCodePoint( codePoint ) == false ) {
+            // throw new IllegalArgumentException( "Invalid codepoint '" + String.format("0x%04X", codePoint) + "'." );
```
jeremiahjstacey, line 98: This appears to be unused, or commented for debugging purposes.  Please remove or adjust as intended.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+
+        // Catch all. The escaped character sequence was invalid.
+        input.reset();
+        throw new IllegalArgumentException( errorMessage );
+    }
+
+    protected int charToCodepoint( Character ch ) {
+
+        final String s = Character.toString(ch);
+        assert (s.length() == 1) : "Ooops";
```
jeremiahjstacey, line 222: Asserts are enabled by the Surefire plugin by default, so if this triggers we will see the output in the test dump.  Would you consider adding more context to the assert failure output to help identify this as the cause when inspecting from a terminal?

EG `"JSONCodec - Unable to convert specified character to codepoint.  Character String conversion must contain more than one character"`

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
```diff
@@ -0,0 +1,242 @@
...
+     * null character is returned.
+     *
+     * @param input a character sequence to decode
+     * @return the decoded version of the encoded character starting at index,
+     *     or null otherwise
+     *
+     * @throws IllegalArgumentException
+     *     if an invalid character sequence is encountered
+     */
+    public Integer decodeCharacter( PushbackSequence<Integer> input )
```
jeremiahjstacey, line 157: The logic flow of this method feels strange to me.

First thing catching my eye is the use of the IllegalArgumentExeption on line 175 to jump to line 209.  I think the try/catch block should be refactored into an if statement.

I appreciate the value of the fast-return on 165, but I have concerns about the other 10 branches that can be used to escape this method.  
 - 8 for each case in the switch
  - 1 at the end for an IllegalArgumentException
  - 1 and the last on line 201.   
   
  If it is possible to reduce this to 3 exit cases I think it will be easier to maintain: 
  - 1 : fast return
  - 2 : Exception case
  - 3 : standard return

I may be able to offer a snippet this evening to assist with this, if that would be helpful.

jeremiahjstacey:
@noloader, I scanned the existing bugs and I don't see a task to associate this PR to. (I may have missed it, sorry)

If something does not exist already, would you please create an issue in the project that encapsulate the intent and value of this change?   Once identified or created, please associate it with this PR for project traceability.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
xeno6696: Also, it's problematic using "UTF" language with the Character class.  Char is from 0-65535.  The whole point of creating the AbstractIntegerCodec was to extend to the full UTF range and traditionally this library pretty much used char/Character as shorthand for Latin-1/ASCII as the developers saw fit.  If this needs to be kept that comment needs to be more reflective of this.

xeno6696: Note that part of this problem is from Java's own use of "Character" to check codepoints.  At any rate it would be better to state that any character whose codepoint is > 0xffff is guaranteed to have an indeterminate value.  Use AbstractIntegerCodec if you want to handle higher values.

noloader: Thanks @xeno6696.

I added you as a collaborator at https://github.com/noloader/esapi-java-legacy. Feel free to update the comments as you see fit.

kwwall:
I believe there is an old, closed GitHub issue for this that was closed
because he thought lost the code when he did a rebase, but he was able to
recover it. We should find that old GitHub issue and just reopen it.

-kevin

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
jeremiahjstacey: I have it down to 4 cases, which may be the minimum:
 1 - First character is null (fast return/no Exception)
 2 - First character is not null, Second character is Null (fast return w/Exception)
 3 - First & Second not null, but no decode can be performed (Exception)
 4 - First & Second not null, decode performed and returned (normal response)

All tests are passing and I'll push to your branch shortly.  I appreciate the consideration and opportunity to contribute to the effort.

noloader: The updated `decodeCharacter` looks great. Awesome job.

noloader:
@jeremiahjstacey 

> I scanned the existing bugs and I don't see a task to associate this PR to. (I may have missed it, sorry)
> 
> If something does not exist already, would you please create an issue in the project that encapsulate the intent and value of this change? Once identified or created, please associate it with this PR for project traceability.

Its kind of a winding road to get here.

This started with Kevin and I talking offline because we need a `JSONEncoder` at `$dayjob`. We did not want to use a JavaScript encoder because the grammars are slightly different. Then, the discussion moved to the mailing list at **_JSON codec discussion on esapi-github_**, https://groups.google.com/a/owasp.org/g/esapi-project-users/c/Q0LTNGxbTQc/m/xoyjMNwCAgAJ .

The mailing list message points to the dead PR at https://github.com/ESAPI/esapi-java-legacy/pull/722 . The original PR is dead because I accidentally blew away the code changes on a rebase. I was able to recover the deleted changes and files through Git's `reflog`, which eventually lead to this [new] PR.

kwwall:
I looked at the issue history and didn't find one. No worries.  I will
create one for you tonight, and reference this PR so they will get linked
together.

-kevin

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
jeremiahjstacey: I don't feel strongly about this.  There is a lot of output in the tests, so the likelihood is that the stacktrace from any assertion failure will be more useful than the message itself.  This will get us in the right area, and I think most developers will be able to fill in the rest with the logic being applied.

Resolving this review thread.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
jeremiahjstacey: @xeno6696, tell me what you'd like this to say and I will update the baseline. I'd like to get this PR merged so I can pull it into the taglib branch so that can also be merged cleanly.

xeno6696: My apologies for the delay, had family in town and also managed to get sick too.  

After some thought I would change it to:

```
// WARNING!! Character based Codecs will only handle the byte range of 0-65535 (0x0-0xffff).  
// Passing any data represented by a higher numerical value will result in a downcast thus
// destroying the original data with undefined results.  

```

noloader: Sounds good. I lifted the original warning from HTMLEncoder, so the change should be made in several places.

jeremiahjstacey: I intend to apply this update before the end of the week, unless someone else beats me to it.

Using the text identified by @xeno6696, in all locations identified by @noloader

jeremiahjstacey: Upon further investigation, there are only 2 places in the baseline that use the block identified by @xeno6696 
```
        // WARNING!! Character based Codecs will silently transform code points that are
        // not legal UTF code points into garbage data as they will cast them to chars.
```

This occurs in the javadoc for `AbstractCodec.encode`
```
  /**
     * {@inheritDoc}
     * </p><p>
     * <b>WARNING!!</b>  {@code Character} based {@code Codec}s will silently transform code points that are not
     * legal UTF code points into garbage data as they will cast them to {@code char}s.
     * Also, if you are implementing an {@code Integer} based codec, these will be silently discarded
     * based on the return from {@code Character.isValidCodePoint( int )}.  This is the preferred
     * behavior moving forward.
     */
    @Override
    public String encode(char[] immune, String input) {
```
and in the `JSONCodec.encodeCharacter` inline comments identified by this review.

The other listed locations do reference the 'silent discard' behavior; however, the only concern being identified is the terminology used around the description of "invalid UTF 8 code points"

For this reason, I will only be updating the two baseline locations:
- Javadoc for `AbstractCodec.encode`
- Comment section in `JSONCodec.encodeCharacter`

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
kwwall: In https://github.com/ESAPI/esapi-java-legacy/pull/753#discussion_r1005491969, @jeremiahjstacey wrote:
> Asserts are enabled by the Surefire plugin by default, so if this triggers we will see the output in the test dump. Would you consider adding more context to the assert failure output to help identify this as the cause when inspecting from a terminal?

That still doesn't mean that we should rely on 'asserts'. The are not generally enabled in runtime by default and even if we are doing so in our Surefire plugin, I would feel better if anything that can be caused by a user accidentally providing incorrect input or not using the classes as per their Javadoc, should be an actual runtime check and not rely on `assert`. Rather, there should be an explicit check and if input is amok, then throw an `IllegalArgumentException`. In other words, do **NOT** use `assert` to check for preconditions. Using `asserts` for postconditions or invariants, I think is okay, as that generally means a program error on our part (assume we've validated all the inputs for sanity), but don't use them for precondition checking.

So, @noloader, if this is a precondition check using assertions, could you change it to an explicit check? Thanks. (I haven't checked, as I'm talking to my mother and trying to listen to her and type at the same time.)

kwwall:
Call for final comments and/ commits on this PR. I'd like to get this and @jeremiahjstacey's PR #756 merged soon so we can do another release. I know Dave Wichers is planning on a AntiSamy point release to fix a CVE in a dependency so we should release when AntiSamy does to prevent people from claiming ESAPI is vulnerable.

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
jeremiahjstacey: @kwwall per your comment on the assert statement I've unresolved that conversation.  Suggest waiting to merge pending updated resolution.

noloader:
On Wed, Nov 16, 2022 at 6:07 AM jeremiahjstacey ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/main/java/org/owasp/esapi/codecs/JSONCodec.java
> <https://github.com/ESAPI/esapi-java-legacy/pull/753#discussion_r1023853508>
> :
>
> > +            // Do nothing. Fall into throw below.
> +        }
> +        catch (IllegalArgumentException e)
> +        {
> +            // Do nothing. Fall into throw below.
> +        }
> +
> +        // Catch all. The escaped character sequence was invalid.
> +        input.reset();
> +        throw new IllegalArgumentException( errorMessage );
> +    }
> +
> +    protected int charToCodepoint( Character ch ) {
> +
> +        final String s = Character.toString(ch);
> +        assert (s.length() == 1) : "Ooops";
>
> @kwwall <https://github.com/kwwall> per your comment on the assert
> statement I've unresolved that conversation. Suggest waiting to merge
> pending updated resolution.
>
The code does not depend on behavior from the assert. The idea behind the
assert is to alert the developer when the code is being debugged. That's
all I wanted to do there: "Hey! Something looks sideways. Take a look at
this!".

In C/C++/ObjC apps, an assert should snap the debugger. In release builds
the assert should be removed. If my understanding is mistaken or a web app
does not alert the developer during a debug session, then would someone
correct it, please.

Now the behavior when asserts are not in effect may be wrong. The behavior
is to return the first character in the pair. I feel like that's not quite
correct. I think the docs say an empty string will be returned on encoding
and decoding errors. Maybe that code path should return "" instead.

kwwall:
In Java, assertions are disabled by default.  We enable them via the
Surefire plugin for our JUnit tests,  but for others development where
ESAPI is used, they would be disabled. So, just consider that as a
possibility.

-kevin

xeno6696:
@kwwall i know I need to review this, I’m on travel for work (again) and I can’t sit down with this until this weekend.  

If we need to push faster than that I’ll trust the group judgment here.

xeno6696:
Given that this was recovered code from #722 I feel like I beat that up pretty well already.  @kwwall I'm okay with moving forward.  I would normally just merge here but it sounded like you had some changes and I don't want to step on your toes here.

xeno6696:
I did want to call myself out to building some unit tests that handled the full Unicode range.  If we stayed with the AbstractCharacterCodec  I have a minor frowny-face just because it was only supposed to be a bridge until all the codecs were converted to Integer-based codecs... however this is my fault as I should have marked that abstract class as deprecated.  (It could be that there are some codecs where for some reason, it makes zero sense at all to expand to handle the full UTF-8 range, but I figured we'd handle those case by case.)

src/main/java/org/owasp/esapi/codecs/JSONCodec.java
kwwall: See my previous comment here. If Unicode characters can extend from 0x0 to 0x10ffff (rather than 0xffff), is there any concern that this could result in a string whose length is >1?

kwwall (commented):
Would like a response to my comment before I approve merging this PR. Changes may or may not be required. I just want to be sure our assumptions about code points are still legitimate as @xeno6696 did the `AbstractCodec` work many years ago when we were using Java 7 and that may have been an older version of Unicode.

src/main/java/org/owasp/esapi/codecs/AbstractCodec.java
```diff
@@ -58,8 +58,10 @@ public AbstractCodec() {
     /**
      * {@inheritDoc}
      * </p><p>
-     * <b>WARNING!!</b>  {@code Character} based {@code Codec}s will silently transform code points that are not
-     * legal UTF code points into garbage data as they will cast them to {@code char}s.
+     * <b>WARNING!!</b>  {@code Character} based {@code Codec}s will only handle the byte range of
+     * 0-65535 (0x0-0xffff).  Passing any data represented by a higher numerical value will result in
```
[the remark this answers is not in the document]
xeno6696: Oy.  So much for a quiet morning!

Okay, so I made a mistake going all the way back to 2017.  Because of the issue I was resolving involving non-BMP characters,  I created the AbstractIntegerCodec to be able to handle the complete UTF-8 range (and future proof it to boot.)  

My underlying assumption all this time was that because of autoboxing, that the `Character` class was 1:1 with `char`.  `char`'s range is 0x00-0xffff.  https://docs.oracle.com/javase/tutorial/java///nutsandbolts/datatypes.html

*this* library before I created the abstract classes was using `char` and `char` alone without using the character classes.  So my framing of the problem began here and is situated with not having read the docs on the Character class, believing it to be ultimately 1:1 with the `char` datatype.  

THAT all being said, I was fully in possession of the fact that under the hood, characters in Java are UTF-16 encoded bytes, and that hidden layer never disappears (and makes you constantly second-guess your code.)  

The end result of all of this is that we support the full range of that java `Character` class documentation.  We haven't necessarily tested the entire character set and I've had a sticky note to do that since 2017.  

The documentation should be updated to reflect what's in the `Character` class.   I still have full faith in the AbstractInteger implementation but the AbstractCharacter class was never tested beyond ensuring that regression tests passed.  Which absolutely means that nothing beyond 0xffff was ever tested.  (Supported, but not tested, and I'm sorry for the error.)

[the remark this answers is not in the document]
kwwall: @xeno6696 - Okay then. That's not as bad as I thought. Seems like everything should work. However, if you haven't done so already, could you create a GitHub issue to outline specifically what needs tested. It seems conceptually simple enough that we would even label it as "Good First Issue", but I don't want it to fall through the cracks for another 5 years or so.

In the meantime, I'm going to merge this PR.

PR #771, opened 2023-01-24
Fix fluido skin plugin version inheritance from pom to src/site/site.xml

davewichers:
Fix fluido skin version inheritance from pom to src/site/site.xml file. And upgrade a number of plugins and maybe one library. Fix some used undeclared and unused declared imports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /target
 /.settings/**
 .classpath
+.java-version
 .project
 *.swp
 *~
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,7 @@
         <version.powermock>2.0.9</version.powermock>
         <version.spotbugs>4.7.3</version.spotbugs>
         <version.spotbugs.maven>4.7.3.0</version.spotbugs.maven>
-        <version.surefire>3.0.0-M8</version.surefire>
+        <version.surefire>3.0.0-M9</version.surefire>
         <project.java.target>1.8</project.java.target>
             <!-- TODO: Be sure to update. Should be date of previous official release -->
             <!-- Exact date in the form 'yyyy-dd-yy 00:00:00' should be used. You can find the previous release date -->
@@ -227,7 +227,7 @@
         <dependency>
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
-            <version>1.4</version>
+            <version>1.5</version>
             <exclusions>
                 <!-- excluded because we directly import newer version. -->
                 <exclusion>
@@ -306,8 +306,14 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+            <version>1.70</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest</artifactId>
+            <artifactId>hamcrest-core</artifactId>
             <version>2.2</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -304,15 +305,9 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
```
kwwall, line 307: This BouncyCastle dependency needs to remain. It is used in for testing the crypto code that allows switching to a different JCE provider. The JUnit tests is created to skip the test if there's no BC library present, but we generally want those tests to run. 
Also, IIRC, this was the most recent version that work when we were running against Java 7. It's possible that a newer version may work with Java 8, but given that this is only used for testing, using an older version is not a big deal.

pom.xml
```diff
@@ -304,15 +305,9 @@
...
-            <groupId>org.bouncycastle</groupId>
-            <artifactId>bcprov-jdk15on</artifactId>
-            <version>1.70</version>
-            <scope>test</scope>
-        </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-core</artifactId>
```
kwwall, line 315: Unless you are 100% certain that hamcrest and hamcrest-core are identical and always will be, let's just leave this as-is too. hamcrest-core is legit (see https://mvnrepository.com/artifact/org.hamcrest/hamcrest-core/2.2). It just seemed like 'hamcrest-core' would be minimal core classes and 'hamcrest' would be that and extra stuff that we are not using.

pom.xml
```diff
@@ -411,7 +406,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-dependency-plugin</artifactId>
-                    <version>3.3.0</version>
```
kwwall, line 414: My typical modus operandi is to update all the plugins and library versions as part of the process of doing a release. (It's even documented that way in the [ESAPI Release Steps](https://github.com/ESAPI/esapi-java-legacy/blob/develop/documentation/ESAPI-release-steps.pdf).) That saves work for us because otherwise we are updating the pom.xml multiple times instead of just once.  And in the past, Dependabot, Snyk, etc. have created PRs that caused the tests to fail. [You'd think they would try to at least run the tests before submitting the PR, but apparently do not. Sigh.]

Anyway, I am okay with _you_  doing this, but I really don't want to set a precedent, because it just gets to be a distraction from the main point of the code review. So, for this particular case, I don't want you to undo things (except where otherwise noted), but I'd recommend skipping this in the future.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /target
 /.settings/**
 .classpath
+.java-version
```
kwwall, line 5: Just curios, what is '.java-version'? I can't say that I've ever seen the file.

Otherwise, I'm okay with the change.

kwwall (changes requested):
See the comments. The removed BC dependency needs to be restored. The hamcrest-core is legit and maybe preferred unless there is some reason to believe 'hamcrest' will always and only refer to the core Hamcrest libraries.

.gitignore
davewichers: .java-version is an artifact of the jenv tool, which allows you to manage/switch Java versions per directory. I use it all the time and it's awesome. You can trivially switch between Java 7, 8, 11, 16 and retest to make sure you are achieving your compatibility goals, etc.

pom.xml
davewichers: Restored

pom.xml
davewichers: Restored

kwwall (approved):

kwwall:
@davewichers - I'll let this fly because I trust you, but from now on, please sign all the commits that you do for ESAPI PRs so I won't have to bypass the branch protection that set up when I do the merge. I've noticed that you had been signing some commits, especially on AntiSamy, but they recently stopped, so either you've been replaced with your evil twin doppelganger or your doing your commits from a machine where you haven't copied your GitHub signing key pair to, ;-)

Issue #767, opened 2023-01-12
Add support for jakarta.servlet package


guadgarcia:
We are in the process of upgrading our Spring Boot application to use Spring Boot 3.0.1 and we are getting an error related to ESAPI. The issue is happening because ServletResponse is now located in a different package.

Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Handler dispatch failed: **java.lang.NoClassDefFoundError: javax/servlet/ServletResponse**] with root cause","context":"","exception":"java.lang.ClassNotFoundException: javax.servlet.ServletResponse\n\tat java.base/java.net.URLClassLoader.findClass(Unknown Source)\n\tat java.base/java.lang.ClassLoader.loadClass(Unknown Source)\n\tat org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:149)\n\tat java.base/java.lang.ClassLoader.loadClass(Unknown Source)\n\tat java.base/java.lang.Class.forName0(Native Method)\n\tat java.base/java.lang.Class.forName(Unknown Source)\n\tat org.owasp.esapi.util.ObjFactory.loadClassByStringName(ObjFactory.java:158)\n\tat org.owasp.esapi.util.ObjFactory.make(ObjFactory.java:81)\n\tat org.owasp.esapi.ESAPI.httpUtilities(ESAPI.java:123)\n\tat org.owasp.esapi.ESAPI.currentRequest(ESAPI.java:70)\n\tat

xeno6696:
It's been a LONG TIME since I set up a Spring application, but IIRC the issue is that ESAPI codes against the servlet-api but it isn't compiled with it.  I would first ensure that your application dependencies are explicit.  

https://mvnrepository.com/artifact/javax.servlet/javax.servlet-api

Some application servers have their own methods for supplying the servlet-api, I know with WebSphere we deployed it as a web app server library so applications didn't need to include it as their own dependency.  

We Code against the specification only so there is no need for any explicit support for Jakarta, Websphere, etc.  You supply your own implementation.  It is 100% supported, you just need to fix your dependencies.

guadgarcia:
[...]

PR #772, opened 2023-01-26
remove dependency on servlet API (use case where not needed)

JosephWitthuhnTR:
This change deals with issue #767.

It does not attempt to resolve the overall issue with Servlet API version compatibility. What it does is remove the dependency on Servlet API for the use case where a user is trying to use the logger functionality, but has "logServerIP" set to false. When not logging the server IP, there is no reason for this code to depend on Servlet API at all.

This allows us to avoid runtime exceptions if we use this code with the new Jakarta package version of Servlet API, in cases where logServerIP is disabled.

Again, this is not a fix for the overall problem. This is a small change that will let some limited use cases not fail due to the problem (so that those use cases can get by until this problem is solved in a more holistic way).

The change stems from part of this discussion:
https://github.com/ESAPI/esapi-java-legacy/discussions/768#discussioncomment-4672270

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/appender/ServerInfoSupplier.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/ServerInfoSupplier.java
@@ -51,9 +51,11 @@ public class ServerInfoSupplier     // implements Supplier<String>
     public String get() {
         // log server, port, app name, module name -- server:80/app/module
         StringBuilder appInfo = new StringBuilder();
-        HttpServletRequest request = ESAPI.currentRequest();
-        if (request != null && logServerIP) {
-            appInfo.append(request.getLocalAddr()).append(":").append(request.getLocalPort());
+        if (logServerIP) {
+            HttpServletRequest request = ESAPI.currentRequest();
+            if (request != null) {
+                appInfo.append(request.getLocalAddr()).append(":").append(request.getLocalPort());
+            }
         }
         if (logAppName) {
             appInfo.append("/").append(applicationName);
--- a/src/test/java/org/owasp/esapi/logging/appender/ServerInfoSupplierTest.java
+++ b/src/test/java/org/owasp/esapi/logging/appender/ServerInfoSupplierTest.java
@@ -25,14 +25,14 @@ public class ServerInfoSupplierTest {
     private HttpServletRequest request;
 
     @Before
-    public void buildStaticMocks() throws Exception {
+    public void buildStaticMocks() {
         request = mock(HttpServletRequest.class);
         mockStatic(ESAPI.class);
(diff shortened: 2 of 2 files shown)


kwwall (commented):
@JosephWitthuhnTR - This looks okay to me, although I wouldn't get your hopes up too much. You are likely to run into other s/javax.servlet/jakarta.servlet/ issues at some point.

@jeremiahjstacey - I'm just looking at this. It LGTM, but the changes are the code that you wrote, so I'm going to let you approve this PR (or not).

jeremiahjstacey:
I have no concerns with the code or tests as written.

I think the value of this update is more related to the intended content of the next ESAPI release.

@kwwall -- do we plan on addressing the javax/jakarta package updates in any other way for the next ESAPI release?

If we do, then this change does not matter.  The order of the logic isn't going to make a difference if we plan to account for it another way.  It is change for the sake of change, which I tend to question.

OTOH if we want to create a release with this content before any other updates are made then it will provide the value that @JosephWitthuhnTR has identified.

Either way the code **can be** merged, depending on the release plan I'm not sure it **should be** due to the possible overlap with the intended value.

----
@JosephWitthuhnTR  - I do not intend to discount your contribution.  I appreciate the fact that you're calling out in the PR description this is not a long-term fix for the problem, as well as the fact that you are invested enough to make time to offer the changes back to the project.   Thank you for your feedback and contributions to the project.

JosephWitthuhnTR:
@kwwall We've tested our application with this after making this modification to our JAR, and we don't run into any other issues. Again, this is for the limited use case where we are just using the logging part of the ESAPI library and we turn the IP logging off. So I am reasonably sure that this fixes our use case (and also pretty certain that it doesn't address the larger problem). What I can't speak to is how many other users (if any) use the JAR for as limited of a use case as we do. This could either fix it for nobody else, or it could help a bunch of people. I did notice that the other individual who submitted the bug (issue #767 and discussion #768) (@guadgarcia) was running into the exact same stack trace / use case as our team was, so it seems that it may help them too.

@jeremiahjstacey I 100% agree. If a new release with the larger fix is "imminent" or "soon", then there is no reason for this modification (and no reason to merge this PR). If that is still weeks/months off, this change "lets us get by" with a smaller fix in the meantime while we wait for the larger fix.

kwwall (approved):
If @jeremiahjstacey is okay with this, then I will approve it. No promises of how soon we get it into a new release though and this barely scratches the surface of the underlying issue.

kwwall:
@JosephWitthuhnTR - While I approve this, it would be nice if you could sign your commit with your GPG key, If that is too much trouble, I can bypass that, but Ive added it as part of the branch protection so ideally you can sign a commit that would be better.

PR #779, opened 2023-03-07
Upgrade the maven site and related plugins, and other little cleanup.

davewichers:
Upgrade some plugins and 1 dependency. Do a little exclusion cleanup, by dropping unneeded exclusions and adding some comments.


kwwall (approved):
LGTM.

PR #786, opened 2023-05-30
Fix reference

noloader:
The LDAP reference we used was RFC 4515, not RFC 4513


kwwall (approved):
LGTM. I even checked the RFC reference this time. Good catch @noloader.

PR #787, opened 2023-06-06
Migration javax -> jakarta if anyone need it

rjunqueira:
Can help someone

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
-    <version>2.5.3.0-SNAPSHOT</version>
+    <version>2.5.3.0-SNAPSHOT-JAKARTA</version>
     <packaging>jar</packaging>
 
     <distributionManagement>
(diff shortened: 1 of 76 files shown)


kwwall:
If you wish to fork ESAPI and support a esapi-jakarta version, then go for it, but the ESAPI team does not have enough resources to support both `javax.servlet` and` jakarta.servlet` namespaces.

We have discussed this at length at https://github.com/ESAPI/esapi-java-legacy/discussions/768, but simply abandoning all current clients using ESAPI is not something that we are ready to do.

Therefore, I am rejecting this PR for the reasons mentioned above. (Now if you can automate this so that 2 versions are supported via Maven so we don't have twice the ongoing support effort, we are certainly willing to do that.)

rjunqueira:
Hello,

Yes i have read your "big issue". Currently i'll maintain my fork (doing periodic refresh) until i find some time to automatize this think as you want (maybe using github scripts/features instead of maven ... avoiding nightmare)

KatrukOV:
Could you explain please, where you got this dependency from?
```
<groupId>org.apache.commons</groupId>
<artifactId>commons-fileupload2</artifactId>
<version>2.0-SNAPSHOT</version>
```

rjunqueira:
@KatrukOV 

Sorry i have forget to include that : 

<repository>
  <id>apache.snapshots</id>
  <name>Apache Development Snapshot Repository</name> 
  <url>https://repository.apache.org/content/repositories/snapshots/</url>
  <releases>
    <enabled>false</enabled>
  </releases>
  <snapshots>
    <enabled>true</enabled>
  </snapshots>
</repository>

So it's here : https://repository.apache.org/content/repositories/snapshots/org/apache/commons/commons-fileupload2/

kwwall:
@KatrukOV - Not that it matters all that much to me, after all, it's your own fork and all, but as a developer, I never want to have a dependency depend on 'snapshots' of anything. That's fine if you want to live on the bleeding edge, but not of you want to provide something that is will be used by enterprise production-ready software. If it were an emergency patch release and that were the only workaround, then perhaps that would be fine as a temporary thing, but  long term, I would be cautious using any library that would even rely on a RC (release candidate) much less a SNAPSHOT version.

rjunqueira:
@KatrukOV  it's ok as we have fixed a specific snapshot version for us to avoid surprises . 

But the only changes of this version 2.0 is the split to supports javax and jakarta (and it will be API compatible). It's near to be released since months (and now their are continuing moving and modifying poms and md files)

cf https://github.com/apache/commons-fileupload

kwwall:
Sounds like the 2.0-SNAPSHOT is only a temporary thing after which you'll start using a stable release, so it's probably not a problem, although I probably would at least note that as a comment in your pom.xml.

KatrukOV:
I think one also needs to change <project.java.target> to the new version.
In my case, I deactivated some plugins and reports, and this is enough for testing ESAPI with jakarta.
Thanks for the opportunity to try this with jakarta.

PR #790, opened 2023-06-17
Update LDAP encoders

noloader:
This commit updates:

* LDAP encoder documentation;
* LDAP encoding in encodeForDN and encodeForLDAP;
* LDAP tests for encodeForDN and encodeForLDAP;

### DIFF
--- a/src/main/java/org/owasp/esapi/Encoder.java
+++ b/src/main/java/org/owasp/esapi/Encoder.java
@@ -444,15 +444,36 @@ public interface Encoder {
     /**
      * Encode data for use in LDAP queries. Wildcard (*) characters will be encoded.
      *
+     * This encoder operates according to RFC 4515, Section 3. RFC 4515 says the following character ranges
+     * are valid: 0x01-0x27, 0x2B-0x5B and 0x5D-0x7F. Characters outside the ranges are hex encoded, and they
+     * include 0x00 (NUL), 0x28 (LPAREN), 0x29 (RPAREN), 0x2A (ASTERISK), and 0x5C (ESC). The encoder will also
+     * encode 0x2F (FSLASH), which is required by Microsoft Active Directory.
+     *
+     * NB: At ESAPI 2.5.3, {@code encodeForLDAP} began strict conformance with RFC 4515. Characters above 0x7F
+     * are converted to UTF-8, and then the byte sequences are hex encoded according to the RFC.
+     *
      * @param input
      *      the text to encode for LDAP
      *
      * @return input encoded for use in LDAP
+     *
+     * @see <a href="https://www.ietf.org/rfc/rfc4515.txt">RFC 4515, Lightweight Directory Access Protocol
+     * (LDAP): String Representation of Search Filters</a>
+     *
+     * @since ESAPI 1.3
      */
     String encodeForLDAP(String input);
 
     /**
      * Encode data for use in LDAP queries. You have the option whether or not to encode wildcard (*) characters.
+     *
+     * This encoder operates according to RFC 4515, Section 3. RFC 4515 says the following character ranges
+     * are valid: 0x01-0x27, 0x2B-0x5B and 0x5D-0x7F. Characters outside the ranges are hex encoded, and they
+     * include 0x00 (NUL), 0x28 (LPAREN), 0x29 (RPAREN), 0x2A (ASTERISK), and 0x5C (ESC). The encoder will also
+     * encode 0x2F (FSLASH), which is required by Microsoft Active Directory.
+     *
+     * NB: At ESAPI 2.5.3, {@code encodeForLDAP} began strict conformance with RFC 4515. Characters above 0x7F
+     * are converted to UTF-8, and then the byte sequences are hex encoded according to the RFC.
      *
      * @param input
      *      the text to encode for LDAP
@@ -460,16 +481,36 @@ public interface Encoder {
      *      whether or not wildcard (*) characters will be encoded.
      *
      * @return input encoded for use in LDAP
+     *
+     * @see <a href="https://www.ietf.org/rfc/rfc4515.txt">RFC 4515, Lightweight Directory Access Protocol
+     * (LDAP): String Representation of Search Filters</a>
+     *
+     * @since ESAPI 1.3
      */
     String encodeForLDAP(String input, boolean encodeWildcards);
 
     /**
      * Encode data for use in an LDAP distinguished name.
      *
+     * This encoder operates according to RFC 4514, Section 3. RFC 4514 says the following character ranges
+     * are valid: 0x01-0x21, 0x23-0x2A, 0x2D-0x3A, 0x3D, 0x3F-0x5B, 0x5D-0x7F. Characters outside the ranges
+     * are hex encoded, and they include 0x00 (NUL), 0x22 (DQUOTE), 0x2B (PLUS), 0x2C (COMMA),
+     * 0x3B (SEMI), 0x3C (LANGLE), 0x3E (RANGLE) and 0x5C (ESC). The encoder will also encode 0x2F (FSLASH),
+     * which is required by Microsoft Active Directory. The leading and trailing characters in a distinguished
+     * name string will also have 0x20 (SPACE) and 0x23 (SHARP) encoded.
+     *
+     * NB: At ESAPI 2.5.3, {@code encodeForDN} began strict conformance with RFC 4514. Characters above 0x7F
+     * are converted to UTF-8, and then the byte sequences are hex encoded according to the RFC.
+     *
      *  @param input
      *          the text to encode for an LDAP distinguished name
      *
      *  @return input encoded for use in an LDAP distinguished name
+     *
+     *  @see <a href="https://www.ietf.org/rfc/rfc4514.txt">RFC 4514, Lightweight Directory Access Protocol
+     *  (LDAP): String Representation of Distinguished Names</a>
+     *
+     *  @since ESAPI 1.3
      */
     String encodeForDN(String input);
 
--- a/src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
+++ b/src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
@@ -311,6 +311,7 @@ public class DefaultEncoder implements Encoder {
         // According to Microsoft docs [1,2], the forward slash ('/') MUST be escaped.
         // According to RFC 4515 Section 3 [3], the forward slash (and other characters) MAY be escaped.
         // Since Microsoft is a MUST, escape forward slash for all implementations. Also see discussion at [4].
+        // Characters above 0x7F are converted to UTF-8 and then hex encoded in the default case.
         // [1] https://docs.microsoft.com/en-us/windows/win32/adsi/search-filter-syntax
         // [2] https://social.technet.microsoft.com/wiki/contents/articles/5312.active-directory-characters-to-escape.aspx
         // [3] https://tools.ietf.org/search/rfc4515#section-3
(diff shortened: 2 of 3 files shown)


kwwall:
@noloader - Before I start reviewing this code in depth, please summarize why you think that we need a new `Encoder` method, `encodeForLdapSearch` and that we can't just have `encodeForLdap` do that functionality as described in RFC 4515. I don't really see that RFC as offering / suggesting that there are 3 different use cases. I think it only deals with encoding DNs / RDNs and encoding for LDAP search filters.  (The encoding for LDIF is mentioned in a different RFC, but that indeed would be a third if you wish to implement that.) The `Encoder` interface is already bloated IMO, so I'm reluctant to add new methods. Thanks. (And I vaguely recall you asking about this in an earlier email, but I didn't have time to comment on. Sorry about that.)

noloader:
Thanks @kwwall.

>  please summarize why you think that we need a new Encoder method, encodeForLdapSearch and that we can't just have encodeForLdap do that functionality as described in RFC 4515. 

I removed that extra encoder.

I also rebased and squashed against the latest tip of develop.

noloader:
@kwwall,

I've been thinking about the old CSRSS backspace bug. In the old days, it was possible to crash Windows `csrss.exe` with a string like `test\b\b\b\b\b\b\b\`. It would cause an out-of-bounds memory access.

Do you think it would be a good idea to encode whitesapce characters and backspace (except traditional SPACE)?

RFC 4514, Section 4, shows an example of escaping the `CR`:

> CN=Before\0dAfter,DC=example,DC=net

kwwall:
Suggest maybe adding a new property something like:
    # Comma-separated list of hex values that should also be encoded in
additon to those mentioned in RFC 4515:
     # Encode backspace and CR.
Encoder.LDAPSearch.extraCharsToEncode=0x08,0x0d

And then if such a property value is non-empty, you encode those
characters. I figured using hex values is better than using \b or \r
because not all hex values have such shortcuts.

Also, this approach is more general and puts the control back in the hands
of the application developers, where it belongs.  That seems a lot better
to me than hard-coding a few select characters.
.
P.S. - No need to rebase and squash. When I do the merge, I usually pick
"Squash and merge", which has a similar effect.

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -444,32 +444,73 @@ public interface Encoder {
     /**
      * Encode data for use in LDAP queries. Wildcard (*) characters will be encoded.
      *
+     * This encoder operates according to RFC 4515, Section 3. RFC 4515 says the following character ranges
+     * are valid: 0x01-0x27, 0x2B-0x5B and 0x5D-0x7F. Characters outside the ranges are hex encoded, and they
+     * include 0x00 (NUL), 0x28 (LPAREN), 0x29 (RPAREN), 0x2A (ASTERISK), and 0x5C (ESC). The encoder will also
+     * encode 0x2F (FSLASH), which is required by Microsoft Active Directory.
+     *
+     * NB: At ESAPI 2.5.3, {@code encodeForLDAP} began strict conformance with RFC 4515. Characters above 0x7F
```
kwwall, line 452: Suggest changing 'NB:' to 'NOTE:' (or perhaps 'WARNING:' if you think it's really important to mention because it is likely to break someone's code.

Outside of legal notation, NB is not commonly used any longer, at least not like it was 30 years ago or so when I used to see it in comments throughout UNIX code bases. I'm not sure younger developers will know what that NB stands for the Latin _nota bene_, commonly translated as "note well" unless perhaps they took Latin in high school.

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -444,32 +444,73 @@ public interface Encoder {
...
 
     /**
      * Encode data for use in LDAP queries. You have the option whether or not to encode wildcard (*) characters.
      *
+     * This encoder operates according to RFC 4515, Section 3. RFC 4515 says the following character ranges
+     * are valid: 0x01-0x27, 0x2B-0x5B and 0x5D-0x7F. Characters outside the ranges are hex encoded, and they
+     * include 0x00 (NUL), 0x28 (LPAREN), 0x29 (RPAREN), 0x2A (ASTERISK), and 0x5C (ESC). The encoder will also
+     * encode 0x2F (FSLASH), which is required by Microsoft Active Directory.
+     *
+     * NB: At ESAPI 2.5.3, {@code encodeForLDAP} began strict conformance with RFC 4515. Characters above 0x7F
```
kwwall, line 475: Same here. Note that this is only a suggestion, but there's no point in making this obtuse and the most common meanings today for the abbreviation "NB" is no longer _nota bene_.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -343,7 +344,18 @@ public String encodeForLDAP(String input, boolean encodeWildcards) {
...
                 default:
-                    sb.append(c);
+                    if (c >= 0x80) {
+                        try {
+                            final byte[] u = String.valueOf(c).getBytes("UTF-8");
+                            for (byte b : u) {
+                                sb.append(String.format("\\%02x", b));
+                            }
+                        } catch (UnsupportedEncodingException ex) {
+                            // UTF-8 is always supported
```
kwwall, line 354: Suggest changing this to:
```java
    ;    // Ignore; UTF-8 is always supported
```
That way, it makes it a little more obvious of the intent when someone sees a\this (false positive) reported by a SAST tool. (Unfortunately, many of them still report this.)

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -390,7 +405,18 @@ public String encodeForDN(String input) {
...
             default:
-                sb.append(c);
+                if (c >= 0x80) {
+                    try {
+                        final byte[] u = String.valueOf(c).getBytes("UTF-8");
+                        for (byte b : u) {
+                            sb.append(String.format("\\%02x", b));
+                        }
+                    } catch (UnsupportedEncodingException ex) {
+                        // UTF-8 is always supported
```
kwwall, line 415: Ditto.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
kwwall, line 1: If you still intend to add encoding for CR and LF, we will need additional tests for that.

kwwall (approved):
This looks fine, but I left a few minor suggestions.
I'm not going to merge yet though, because I'm not sure if you wanted to do the encoding for CR and LF like you suggested (or, generalizing that, like I recommended -- so it could be disabled) or want to incorporate the suggestions to some of the comments I made here as part of this review. Therefore I will wait to merge. Once you are satisfied with all your changes, just @-me with comment you are ready to merge.

kwwall:
@xeno6696 - If you want to review this PR, that's fine but not strictly necessary. But if you do, please hold off on the merging (assuming that you approve in the code review) until we hear back from @noloader to see if he intends to make further changes. I'd rather make related changes as part of this PR than having to create a new PR.

Also, @noloader - could you please create a GitHub Issue for this and associate it to this PR. It can be something really simple like "Ensure strict conformance with section 3 of RFC 4515 in terms of character ranges" or more detailed, depending on your tastes. That will help me when I prepare the release notes.

kwwall:
@noloader - did you intend to add to this PR to discuss the reply I made in https://github.com/ESAPI/esapi-java-legacy/pull/790#issuecomment-1596484323 or are you through?

@xeno6696 - We need one more person to approve this. Can you look at it?

kwwall:
@xeno6696 or @jeremiahjstacey - We need one more person to approve this. Can one of you PLEASE look at this ASAP. I'd like to get it into the upcoming ESAPI 2.5.3.0 release.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -343,7 +344,18 @@ public String encodeForLDAP(String input, boolean encodeWildcards) {
                     sb.append("\\00");
                     break;
                 default:
-                    sb.append(c);
+                    if (c >= 0x80) {
+                        try {
+                            final byte[] u = String.valueOf(c).getBytes("UTF-8");
```
xeno6696, line 349: Something feels a bit off here.  char is 16bit.... but the data in `input` could be a string of characters that won't live through this translation.  I'm thinking of my unit tests over here:

```
    @Test
    public void test32BitCJK(){
        String s = "𡘾𦴩𥻂";
        String expected = "&#x2163e;&#x26d29;&#x25ec2;";
        String bad = "&#xd845;&#xde3e;&#xd85b;&#xdd29;&#xd857;&#xdec2;";
        assertEquals(false, expected.equals(bad));
        assertEquals(expected, codec.encode(new char[0], s));
    }

    @Test
    public void test32BitCJKMixedWithBmp(){
        String s = "𡘾𦴩<𥻂";
        String expected = "&#x2163e;&#x26d29;&lt;&#x25ec2;";
        String bad = "&#xd845;&#xde3e;&#xd85b;&#xdd29;&#xd857;&#xdec2;";
        assertEquals(false, expected.equals(bad));
        assertEquals(expected, codec.encode(new char[0], s));
    }
```
    
So imagine that the input is `𡘾𦴩𥻂`

I would think that there would be an overflow the second we call `char c = input.charAt(i);`

Because under the hood, java is representing `𡘾𦴩𥻂` as a UTF-16 buffer.  `𡘾` would get translated to garbage and then disaster follows.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
```diff
@@ -390,7 +405,18 @@ public String encodeForDN(String input) {
                 sb.append("\\;");
                 break;
             default:
-                sb.append(c);
+                if (c >= 0x80) {
+                    try {
+                        final byte[] u = String.valueOf(c).getBytes("UTF-8");
```
xeno6696, line 410: Same comment here.

xeno6696 (commented):
@kwwall @noloader If the LDAP standard has been expanded to accomodate UTF-8 encoded bytes then our API for LDAP encoding needs to be broken out into Codecs so that way we can at least maintain legacy functionality by leaving the old behavior alone and then more carefully considering the actual use case of UTF-8 bytes.  

I don't feel comfortable approving this because as written, this directly and permanently affects the encodeForLDAP method.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
kwwall: @xeno6696 - Can this wait until _after_ this upcoming 2.5.3.0 release? In other words, I'll phrase the question this way: Which is less wrong? The way the code presently is or the way this PR is? If you honestly think that the code in this PR is worse than what we presently have, we can exclude it from this next release, but I'm not sure I see it this way. Even if this code has the issues that you mention, I think it moves us toward the direction of the of compliance with the RFC. So, let's not let the perfect become the enemy of the good here. IMO, unless this is going to open up worse security holes in your opinion, I think the way forward should create a GitHub issue as to what needs fixed once this PR is merged and then merge this and deal with the rest of it later.

@noloader, @xeno6696 - Thoughts?

xeno6696:
> @noloader - Before I start reviewing this code in depth, please summarize why you think that we need a new `Encoder` method, `encodeForLdapSearch` and that we can't just have `encodeForLdap` do that functionality as described in RFC 4515. I don't really see that RFC as offering / suggesting that there are 3 different use cases. I think it only deals with encoding DNs / RDNs and encoding for LDAP search filters. (The encoding for LDIF is mentioned in a different RFC, but that indeed would be a third if you wish to implement that.) The `Encoder` interface is already bloated IMO, so I'm reluctant to add new methods. Thanks. (And I vaguely recall you asking about this in an earlier email, but I didn't have time to comment on. Sorry about that.)

I didn't realize that @noloader was implementing your suggestion here.  It's kinda late for me to disagree with it but my strategy would have been to create a new method with the better logic and then followed the deprecation strategy.  I don't see the need to force a block here @kwwall and I think I need to develop a weekly routine to check up on things so I don't hold them up.

src/main/java/org/owasp/esapi/reference/DefaultEncoder.java
xeno6696: My main worry is this:  Assuming that the rules we were breaking were breaking because (like browsers) some LDAP systems don't obey the RFC to the letter anyway, we could unintentionally break something.  That said, I'm also unsure of how many folks are using ESAPI for this anyway.  I won't stand in the way of the merge and I'm okay with punting for future issues, but then you also stated a desire to push 2.x into maintenance only which effectively bars that.  

Just putting that out there.

PR #799, opened 2023-09-08
Using Eclipse Transformer plugin to create Jakarta JAR

jcputney:
This PR adds a Maven profile called "jakarta" that will generate a transformed JAR file, compatible with the `jakarta.servlet` changes for Jakarta EE 9+

I don't know how you public packages, and I doubt this is a full solution, but it at least gets you pointed in the right direction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ bin/
 ciphertext-portable.ser
 ReferenceEncryptedProperties.test.txt
 test.out
+.DS_Store
--- a/pom.xml
+++ b/pom.xml
@@ -253,6 +253,21 @@
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-api</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.apache.xmlgraphics</groupId>
+                    <artifactId>batik-css</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.xmlgraphics</groupId>
+            <artifactId>batik-css</artifactId>
+            <version>1.17</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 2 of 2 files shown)


.gitignore
kwwall, line 1: Okay; I don't think any of us use Macs for ESAPI development, but harmless enough.

kwwall, line 1: I confirmed this by pulling down this PR locally and running 'mvn compile install' and it seems to work and do what I want it to do. @jcputney - Would you take a quick look at our [ESAPI Release Steps document](https://github.com/ESAPI/esapi-java-legacy/blob/develop/documentation/ESAPI-release-steps.pdf) and let me know if you spot anything that we need to add or do differently. I'd rather not get it partway uploaded to Maven Central as the pre-release and then have the release fail. But if you check it and let me know if we need to make any changes, I will try to get this as part of the ESAPI 2.5.3.0 release which I am presently working on.

kwwall (commented):
@jcputney - Looks good and works as expected, but I am concerned about the release details to Maven Central. Please see my 2nd comment associated with our pom.

kwwall:
@xeno6696 and @jeremiahjstacey - I will approve this after my PR #809 gets merged. If it gets merged now, I think it likely will cause merge conflicts.

.gitignore
kwwall: @jcputney - I'm getting close to releasing ESAPI 2.5.3.0. I may include this with it, or I make create a 2.6.0.0 release shortly thereafter, but could you please do a quick read through the ESAPI release notes document that I mentioned in my previous comment to see if that should suffice for this as well? Thanks.

kwwall (approved):
This at least seems to work up through the 'mvn compile install' step. It creates a **esapi-_\<release\>_-jakarta.jar** whose classes have references to the` jakarta.servlet` package namespace rather than the `javax.servlet` namespace. Whether it works when trying to upload to Maven Central, that remains to be seen, but it's worth is shot. If that part doesn't work, I'll back this change out and release 2.5.3.0 without this.

JLLeitschuh:
If this worked, it may be good to go update the README to reflect that this JAR now exists and is published to maven central?

kwwall:
It did work. I mentioned it in the release notes and on the Discussion
#768. But I think noting it in the README.md is also a good idea. Thanks
for the suggestion.

-kevin

PR #807, opened 2023-11-03
fix typo, remove double `and`

robstoll:


kwwall (changes requested):
The same typo needs fixed on line 30 of 'src/test/resources/esapi/validation.properties'. Would you be so kind as to fix that typo as well, and they I shall approve it? Thanks.

Also, I created GitHub issue #808 to track this for the release notes.

kwwall (approved):
LGTM. Thanks.

Issue #811, opened 2023-11-20
Option to omit event type prefix in logs


RodolfoAndre:
**Is your feature request related to a problem? Please describe.**
I am in a process to improve security of an application using ESAPI library. I've noticed that there is an obligatory prefix that, since I have a huge infrastructure, the event type prefix would increase our cost because we use a paid monitoring service.

**Describe the solution you'd like**
It would help us if there was an optional property in ESAPI.properties file to indicate if the event type prefix log should be included or not when printing logs. (Just from the log message)

**Describe alternatives you've considered including other security libraries**
An alternative would be when all properties related to prefix append (logUserInfo, logClientInfo, logServerIp, logApplicationName) and the appName is empty or null, this prefix would not be included.

**Additional context**
I've had a [discussion ](https://github.com/ESAPI/esapi-java-legacy/discussions/810#discussioncomment-7623059) with [kwwall](https://github.com/kwwall) with some ideas. I will fork the repo and send the pull request as an example.

kwwall:
Reference to Discussion #810.

PR #813, opened 2023-11-21
Added property to omit event type information in logs

RodolfoAndre:
This PR adds the property "Logger.OmitEventTypeInLogs" that will make event type information not be prepended in log messages when set to true (default false).

### DIFF
--- a/configuration/esapi/ESAPI.properties
+++ b/configuration/esapi/ESAPI.properties
@@ -400,6 +400,8 @@ Logger.LogServerIP=true
 Logger.UserInfo=true
 # Determines whether ESAPI should log the session id and client IP.
 Logger.ClientInfo=true
+# Determines whether ESAPI should log the event type or not
+Logger.OmitEventTypeInLogs=false
 
 #===========================================================================
 # ESAPI Intrusion Detection
--- a/src/main/java/org/owasp/esapi/PropNames.java
+++ b/src/main/java/org/owasp/esapi/PropNames.java
@@ -111,6 +111,7 @@ public final class PropNames {
     public static final String LOG_ENCODING_REQUIRED                                                            = "Logger.LogEncodingRequired";
     public static final String LOG_APPLICATION_NAME                                                             = "Logger.LogApplicationName";
     public static final String LOG_SERVER_IP                                                                    = "Logger.LogServerIP";
+    public static final String OMIT_EVENT_TYPE_IN_LOGS                                                          = "Logger.OmitEventTypeInLogs";
 
     public static final String VALIDATION_PROPERTIES                                                            = "Validator.ConfigurationFile";
     public static final String VALIDATION_PROPERTIES_MULTIVALUED                                                = "Validator.ConfigurationFile.MultiValued";
--- a/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
@@ -35,6 +35,8 @@ public class LogPrefixAppender implements LogAppender {
     private final boolean logApplicationName;
     /** Application Name to record. */
     private final String appName;
+    /** Whether to omit event type in logs or not. */
+    private final boolean omitEventTypeInLogs;
 
     /**
      * Ctr.
@@ -44,17 +46,23 @@ public class LogPrefixAppender implements LogAppender {
      * @param logServerIp        Whether or not to record server ip information
      * @param logApplicationName Whether or not to record application name
      * @param appName            Application Name to record.
+     * @param omitEventTypeInLogs Application Name to record.
      */
-    public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName) {
+    public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName, boolean omitEventTypeInLogs) {
         this.logUserInfo = logUserInfo;
         this.logClientInfo = logClientInfo;
         this.logServerIp = logServerIp;
         this.logApplicationName = logApplicationName;
         this.appName = appName;
+        this.omitEventTypeInLogs = omitEventTypeInLogs;
     }
 
     @Override
     public String appendTo(String logName, EventType eventType, String message) {
+        if (omitEventTypeInLogs) {
+            return message;
+        }
+
         EventTypeLogSupplier eventTypeSupplier = new EventTypeLogSupplier(eventType);
 
         UserInfoSupplier userInfoSupplier = new UserInfoSupplier();
--- a/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
@@ -15,7 +15,11 @@
 
 package org.owasp.esapi.logging.appender;
 
+import org.owasp.esapi.ESAPI;
 import org.owasp.esapi.Logger.EventType;
+import org.owasp.esapi.errors.ConfigurationException;
+
+import static org.owasp.esapi.PropNames.OMIT_EVENT_TYPE_IN_LOGS;
 
 /**
  * LogAppender Implementation which can prefix the common logger information for
@@ -36,7 +40,17 @@ public class LogPrefixAppender implements LogAppender {
     /** Application Name to record. */
     private final String appName;
     /** Whether to omit event type in logs or not. */
-    private final boolean omitEventTypeInLogs;
+    private static boolean omitEventTypeInLogs;
+
+    static {
+
+        try {
+            omitEventTypeInLogs =
+                    ESAPI.securityConfiguration().getBooleanProp(OMIT_EVENT_TYPE_IN_LOGS);
+        } catch (ConfigurationException ex) {
+            omitEventTypeInLogs = false;
+        }
+    }
 
     /**
      * Ctr.
@@ -46,23 +60,17 @@ public class LogPrefixAppender implements LogAppender {
      * @param logServerIp        Whether or not to record server ip information
      * @param logApplicationName Whether or not to record application name
      * @param appName            Application Name to record.
-     * @param omitEventTypeInLogs Application Name to record.
      */
-    public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName, boolean omitEventTypeInLogs) {
+    public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName) {
         this.logUserInfo = logUserInfo;
         this.logClientInfo = logClientInfo;
         this.logServerIp = logServerIp;
         this.logApplicationName = logApplicationName;
         this.appName = appName;
-        this.omitEventTypeInLogs = omitEventTypeInLogs;
     }
 
     @Override
     public String appendTo(String logName, EventType eventType, String message) {
-        if (omitEventTypeInLogs) {
-            return message;
-        }
-
         EventTypeLogSupplier eventTypeSupplier = new EventTypeLogSupplier(eventType);
 
         UserInfoSupplier userInfoSupplier = new UserInfoSupplier();
@@ -75,7 +83,7 @@ public class LogPrefixAppender implements LogAppender {
         serverInfoSupplier.setLogServerIp(logServerIp);
         serverInfoSupplier.setLogApplicationName(logApplicationName, appName);
 
-        String eventTypeMsg = eventTypeSupplier.get().trim();
+        String eventTypeMsg = omitEventTypeInLogs ? "" : eventTypeSupplier.get().trim();
         String userInfoMsg = userInfoSupplier.get().trim();
         String clientInfoMsg = clientInfoSupplier.get().trim();
         String serverInfoMsg = serverInfoSupplier.get().trim();
@@ -89,7 +97,7 @@ public class LogPrefixAppender implements LogAppender {
         String[] optionalPrefixContent = new String[] {userInfoMsg + clientInfoMsg, serverInfoMsg};
 
         StringBuilder logPrefix = new StringBuilder();
-        //EventType is always appended
+
         logPrefix.append(eventTypeMsg);
 
         for (String element : optionalPrefixContent) {
@@ -99,6 +107,11 @@ public class LogPrefixAppender implements LogAppender {
             }
         }
 
-        return String.format(RESULT_FORMAT, logPrefix.toString(), message);
+        if (logPrefix.toString().trim().isEmpty()) {
+            // if there isn't any log prefix we just send back the message without touching it
+            return message;
+        }
+
+        return String.format(RESULT_FORMAT, logPrefix.toString().trim(), message);
     }
 }
--- a/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
@@ -107,11 +107,7 @@ public class LogPrefixAppender implements LogAppender {
             }
         }
 
-        if (logPrefix.toString().trim().isEmpty()) {
-            // if there isn't any log prefix we just send back the message without touching it
-            return message;
-        }
-
-        return String.format(RESULT_FORMAT, logPrefix.toString().trim(), message);
+        String prefix = logPrefix.toString().trim();
+        return prefix.isEmpty() ? message : String.format(RESULT_FORMAT, prefix, message);
     }
 }
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -41,6 +41,7 @@ import static org.owasp.esapi.PropNames.LOG_CLIENT_INFO;
 import static org.owasp.esapi.PropNames.LOG_APPLICATION_NAME;
 import static org.owasp.esapi.PropNames.APPLICATION_NAME;
 import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
+import static org.owasp.esapi.PropNames.OMIT_EVENT_TYPE_IN_LOGS;
 
 /**
  * LogFactory implementation which creates JAVA supporting Loggers.
@@ -79,7 +80,8 @@ public class JavaLogFactory implements LogFactory {
         boolean logApplicationName = ESAPI.securityConfiguration().getBooleanProp(LOG_APPLICATION_NAME);
         String appName = ESAPI.securityConfiguration().getStringProp(APPLICATION_NAME);
         boolean logServerIp = ESAPI.securityConfiguration().getBooleanProp(LOG_SERVER_IP);
-        JAVA_LOG_APPENDER = createLogAppender(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName);
+        boolean omitEventTypeInLogs = ESAPI.securityConfiguration().getBooleanProp(OMIT_EVENT_TYPE_IN_LOGS);
+        JAVA_LOG_APPENDER = createLogAppender(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName, omitEventTypeInLogs);
 
         Map<Integer, JavaLogLevelHandler> levelLookup = new HashMap<>();
         levelLookup.put(Logger.ALL, JavaLogLevelHandlers.ALWAYS);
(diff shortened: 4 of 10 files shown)


src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -79,7 +80,8 @@ public class JavaLogFactory implements LogFactory {
         boolean logApplicationName = ESAPI.securityConfiguration().getBooleanProp(LOG_APPLICATION_NAME);
         String appName = ESAPI.securityConfiguration().getStringProp(APPLICATION_NAME);
         boolean logServerIp = ESAPI.securityConfiguration().getBooleanProp(LOG_SERVER_IP);
-        JAVA_LOG_APPENDER = createLogAppender(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName);
+        boolean omitEventTypeInLogs = ESAPI.securityConfiguration().getBooleanProp(OMIT_EVENT_TYPE_IN_LOGS);
```
kwwall, line 83: For reasons noted in a comment in the LogPrefixAppender CTOR that you modifed, there is a (probably good) possibility that this property will not be found at all in the **ESAPI.properties** file, so you need to be prepared to catch `ConfigurationException` here and set `omitEventTypeInlogs` to `false` should that happen. Because we can't let a `ConfigurationException` propagate back to the caller if **Logger.OmitEventTypeInLogs** is not found. That would result in breaking people's code and as a result is just not acceptable.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -156,11 +158,12 @@ public class JavaLogFactory implements LogFactory {
...
      * @param logServerIp
      * @param logClientInfo
+     * @param omitEventTypeInLogs
      *
      * @return LogAppender instance.
      */
-    /*package*/ static LogAppender createLogAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName) {
-        return new LogPrefixAppender(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName);
+    /*package*/ static LogAppender createLogAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName, boolean omitEventTypeInLogs) {
+        return new LogPrefixAppender(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName, omitEventTypeInLogs);
```
kwwall, line 166: May wish to change this, depending  @jeremiahjstacey's take on my comments on the `LogPrefixAppender` CTOR. But at even if this remains the same here, we can't simple remove the old CTOR as the CTOR is public and thus off limits. (Once we start using Java 9 or later, we can better control this with Java modules, but until then, no changes to signatures of public or protected methods unless it is absolutely required to address a vulnerability.)

src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogFactory.java
```diff
@@ -69,7 +71,8 @@ public class Slf4JLogFactory implements LogFactory {
         boolean logApplicationName = ESAPI.securityConfiguration().getBooleanProp(LOG_APPLICATION_NAME);
         String appName = ESAPI.securityConfiguration().getStringProp(APPLICATION_NAME);
         boolean logServerIp = ESAPI.securityConfiguration().getBooleanProp(LOG_SERVER_IP);
-        SLF4J_LOG_APPENDER = createLogAppender(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName);
+        boolean omitEventTypeInLogs = ESAPI.securityConfiguration().getBooleanProp(OMIT_EVENT_TYPE_IN_LOGS);
```
kwwall, line 74: Same comment here as for `JavaLogFactory` class... for reasons noted in a comment in the LogPrefixAppender CTOR that you modifed, there is a (probably good) possibility that this property will not be found at all in the **ESAPI.properties** file, so you need to be prepared to catch `ConfigurationException` here and set `omitEventTypeInlogs` to `false` should that happen. Because we can't let a `ConfigurationException` propagate back to the caller if **Logger.OmitEventTypeInLogs** is not found. That would result in breaking people's code and as a result is just not acceptable.

kwwall (changes requested):
@RodolfoAndre - Please kindly address the code review comments. You may want to wait until @jeremiahjstacey adds his comments as he is the author and expert of the revised ESAPI logging.

kwwall:
@jeremiahjstacey - Please code review this when you have a moment. I have already made a lot of comments, but I'm not sure I understand the JUnit tests well enough unless I really dive into the gritty details of how they work.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
```diff
@@ -44,17 +46,23 @@ public class LogPrefixAppender implements LogAppender {
      * @param logServerIp        Whether or not to record server ip information
      * @param logApplicationName Whether or not to record application name
      * @param appName            Application Name to record.
+     * @param omitEventTypeInLogs Application Name to record.
      */
-    public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName) {
+    public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName, boolean omitEventTypeInLogs) {
```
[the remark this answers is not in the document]
RodolfoAndre: Ok, I have changed back. I have a question where should I get the OmitEventTypeInLogs property? 

I could create a static variable in Slf4JLogFactory and use the static method to get this property and create a setOmitEventTypeInLogs method in LogPrefixAppender to set the value;
I can do that when the LogPrefixAppender instanced in Slf4JLogFactory static method. So I would get the property and create a setOmitEventTypeInLogs method in LogPrefixAppender to set the value.

Which approach do you think would be better or do you have a suggestion?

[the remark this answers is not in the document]
kwwall: ESAPI generally makes the assumption (except when running JUnit tests) that the value of a a specific property provided in ESAPI.properties doesn't change in the middle of your application, so as a result, it's generally fine to assume one a property is set, you can store its value in a static class variable.

That said, what I was thinking in the context of LogPrefixAppender was rather than changing the CTOR (or adding a new one), change your declaration of:
```java
    private final boolean omitEventTypeInLogs;
```
to
```java
    import static org.owasp.esapi.PropNames.OMIT_EVENT_TYPE_IN_LOGS;
    ...
    private static boolean omitEventTypeInLogs;

    static {
        try {
            omitEventTypeInLogs =
                ESAPI.securityConfiguration().getBooleanProp( OMIT_EVENT_TYPE_IN_LOGS );
        } catch (Exception ex) {    // Usually this will be ConfigurationException
            omitEventTypeInLogs = false;
        }
    }
```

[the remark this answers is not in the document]
RodolfoAndre: Yes! thank you. With this solution most of the comments bellow are fulfilled. I will send another patch shortly.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
```diff
@@ -44,17 +46,23 @@ public class LogPrefixAppender implements LogAppender {
...
         this.logClientInfo = logClientInfo;
         this.logServerIp = logServerIp;
         this.logApplicationName = logApplicationName;
         this.appName = appName;
+        this.omitEventTypeInLogs = omitEventTypeInLogs;
     }
 
     @Override
     public String appendTo(String logName, EventType eventType, String message) {
+        if (omitEventTypeInLogs) {
```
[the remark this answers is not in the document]
RodolfoAndre: Yes, I was thinking in a pespective that everything inside the [] was related to event type. So I have made the change you suggested and then it will check if the prefix is empty, if so it will return just the message without the prefix, otherwise it will add the prefix in the message.

configuration/esapi/ESAPI.properties
jeremiahjstacey, line 1: I believe there are also ESAPI.properties files in the test scope that should also be updated
https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/test/resources/esapi/ESAPI.properties

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
```diff
@@ -35,6 +39,18 @@ public class LogPrefixAppender implements LogAppender {
...
+    private static boolean omitEventTypeInLogs;
+
+    static {
+
+        try {
+            omitEventTypeInLogs =
+                    ESAPI.securityConfiguration().getBooleanProp(OMIT_EVENT_TYPE_IN_LOGS);
+        } catch (ConfigurationException ex) {
+            omitEventTypeInLogs = false;
+        }
```
jeremiahjstacey, line 52: ConfigurationException should not be caught.

Additionally, doing this in the static block may prevent a misconfigured application from correctly propagating the runtime exception. 

I realize that the static keyword is intended to optimize the number of instances of this variable, but this project really does have enough static content.  Let the variable die with the instance.

Please consider:
1.  Removing the catch from the initialization -- that exception should escape if it happens (See DefaultSecurityProvider.getBooleanProp)
2. make the omitEventTypeInLogs a member varable of the class (remove static modifier), and move the initialization to the constructor.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
```diff
@@ -91,6 +107,11 @@ public String appendTo(String logName, EventType eventType, String message) {
             }
         }
 
-        return String.format(RESULT_FORMAT, logPrefix.toString(), message);
+        if (logPrefix.toString().trim().isEmpty()) {
+            // if there isn't any log prefix we just send back the message without touching it
+            return message;
+        }
+
+        return String.format(RESULT_FORMAT, logPrefix.toString().trim(), message);
```
jeremiahjstacey, line 115: I would feel better about this if it was a single return as a ternary:

```
String prefix = logPrefix.toString().trim();
return prefix.isEmpty() ? message : String.format(RESULT_FORMAT, prefix, message);
```

src/test/java/org/owasp/esapi/logging/appender/LogPrefixAppenderTest.java
```diff
@@ -159,8 +186,26 @@ private void runTest(String typeResult, String userResult, String clientResult,
 
         //Since everything is mocked these booleans don't much matter aside from the later verifies
         LogPrefixAppender lpa = new LogPrefixAppender(false, false, false, false, null);
-        String result =   lpa.appendTo(testLoggerName, testEventType, testLogMessage);
 
-        assertEquals(exResult + " " + testName.getMethodName() + "-MESSAGE", result);
+        // Using reflection API to set omitEventTypeInLogs field in LogPrefixAppender.
+        setOmitEventTypeInLogsFieldUsingReflection(lpa, omitEventTypeInLogs);
```
jeremiahjstacey, line 191: From my previous comments about making the variable a member instance, we will not need reflection here.
```
System.setProperty(PropNames.OMIT_EVENT_TYPE_IN_LOGS, omitEventTypeInLogs);

 //Since everything is mocked these booleans don't much matter aside from the later verifies
 LogPrefixAppender lpa = new LogPrefixAppender(false, false, false, false, null);
```

src/main/java/org/owasp/esapi/PropNames.java
```diff
@@ -111,6 +111,7 @@ public final class PropNames {
     public static final String LOG_ENCODING_REQUIRED                                                            = "Logger.LogEncodingRequired";
     public static final String LOG_APPLICATION_NAME                                                             = "Logger.LogApplicationName";
     public static final String LOG_SERVER_IP                                                                    = "Logger.LogServerIP";
+    public static final String OMIT_EVENT_TYPE_IN_LOGS                                                          = "Logger.OmitEventTypeInLogs";
```
jeremiahjstacey, line 114: Convention consistency:
Suggest following the existing naming scheme:
`LOG_EVENT_TYPE`

Maintains consistency with the positive action verbiage used in the other properties.

This will likely mean you'll need to change all of the booleans in the tests as well, but it will be more consistent with the existing convention.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
jeremiahjstacey: I see @kwwall  and I are in disagreement about the ConfigurationException.   Please disregard my  suggestion and follow his guidance.

I believe moving the initialization to the constructor and making it non-static is still valid.

Apologies for not being more dilligent about reading the previous conversation before jumping into the code.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: :(  Catching a runtime exception to avoid breaking feels dirty to me, but I do agree it has the unfavorable result of breaking implementations when new properties are required.

It is worth noting that there are actually 2 cases that the Exception may occur:
1.  the value is not set
2. the value is not of the expected type (cannot be converted to a boolean)

In the second case, catching this exception may result in hiding a problem with a property value.  I do think it is a less likely scenario, but still worth noting.

```
 @Override
    public Boolean getBooleanProp(String propertyName) throws ConfigurationException {
        try {
            return esapiPropertyManager.getBooleanProp(propertyName);
        } catch (ConfigurationException ex) {
            String property = properties.getProperty( propertyName );
            if ( property == null ) {
                throw new ConfigurationException( "SecurityConfiguration for " + propertyName + " not found in ESAPI.properties");
            }
            if ( property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes" ) ) {
                return true;
            }
            if ( property.equalsIgnoreCase("false") || property.equalsIgnoreCase( "no" ) ) {
                return false;
            }
            throw new ConfigurationException( "SecurityConfiguration for " + propertyName + " has incorrect " +
                    "type");
        }
    }
```

kwwall: @jeremiahjstacey - True, but the more common error is just going to be that people don't update their ESAPI.properties file regularly, so the property would be missing and I don't want ESAPI to start failing (and thus breaking code) because of that. We could have the `ConfigurationException` logged via `logSpecial()`, but I'd prefer dirty over adding a new CTOR and all the other ugliness that brings into ESAPI code.

configuration/esapi/ESAPI.properties
kwwall: Absolutely correct. I thought at the time that I originally reviewed this, he had updated the **src/test/resources/esapi/ESAPI.properties** file to add that new property. Maybe he undid it in the last commit or maybe I just was imagining it, but yeah, it needs to be there too.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
kwwall: My reason for making it static and putting it in a static initializer goes as follows:

1. There are potentially a LOT of uses of ESAPI logging statements so the `LogPrefixAppender` CTOR will be called many times.
2. It seems likely that many applications will not have the new `Logger.OmitEventTypeInLogs` property.
3. If that property is missing, as previously discussed, it will cause a `ConfigurationException` to be thrown each time.
4. Throwing / catching exceptions is a relatively expensive operation from a Java execution perspective.
5. Overall, this could make the effect of logging considerably slower.
6. If we make it a class variable instead and place it in a static initializer as I suggested, we only suffer that throw/catch overhead once.
7. We generally assume that the value of an ESAPI property does not change in the course of a running application, so given that assumption, this is a sound, although rather unorthodox approach.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
kwwall: I'm good with that.

src/test/java/org/owasp/esapi/logging/appender/LogPrefixAppenderTest.java
kwwall: Except, setting this is a `System` property **_still_** won't make that property get picked up by `ESAPI.securityConfiguration().getBooleanProp()`.

@RodolfoAndre - We have a "trick" to make ESAPI properties work with JUnit tests so we can simulate various values that @jeremiahjstacey may have forgotten about. We use the `org.owasp.esapi.SecurityConfigurationWrapper` class for that. Rather than trying to describe how to use that for JUnit tests, I will instead refer you to
https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/test/java/org/owasp/esapi/reference/validation/HTMLValidationRuleCleanTest.java
as a typical example of how we use it. If after looking at that, you still can't figure it out, then let me know.

src/main/java/org/owasp/esapi/PropNames.java
kwwall: The static `String` is generally based on the last portion of the property name, so I think he is consistent given the name of the property. (And I suggested the property name in a GitHub issue #811 and in the discuss that it references (#810). So, my bad for that.

I'm fine if we want to rename the property to match the rest of these **Logger**  related properties, but I will leave it to you to suggest a property name them. I figured that if we called it the more conventional **Logger.LogOmitEventType** that developers may incorrectly infer from that that they could omit `Logger.EVENT_TYPE` when invoking the logger and **Logger.LogOmitEventTypeInLogs** almost sounded redundant, so I just dropped the initial 'Log' when I proposed a name. I do think this is consistent here given how the property is currently named though, so I suggest we leave this as-is here or else change the property name. And, as mentioned, I'm fine with either of you proposing a different property name.

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
jeremiahjstacey: I appreciate the explanation and understand the intention.  Thank you for the consideration.

Resolving conversation, no changes requested/expected.

src/test/java/org/owasp/esapi/logging/appender/LogPrefixAppenderTest.java
jeremiahjstacey: No changes needed here as a result of no changes from previous conversations.

jeremiahjstacey: My initial context is invalid for removing reflection utility, but if we're able to do that I think it would be a good thing.
Thank you for the context @kwwall

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
RodolfoAndre: We can incorporate the check of property value as you mentitoned and ignore it if the property is not present. What do you think?

`if (property != null) {
                if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
                    omitEventTypeInLogs = true;
                } else if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
                    omitEventTypeInLogs = false;
                } else if (property != null) {
                    throw new ConfigurationException("SecurityConfiguration for " + OMIT_EVENT_TYPE_IN_LOGS + " has incorrect " +
                            "type");
                }
            }`

src/test/java/org/owasp/esapi/logging/appender/LogPrefixAppenderTest.java
RodolfoAndre: @kwwall I could change the property just once. After the first initalization the class won't get the property again. So I can't have test for true and false in the same class.

 It will work if I change the test to another file or create a static class inside this one just for testing the flag. Do you have any idea what may I do?

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: The code I posted is the default implementation of getBooleanProp from the DefaultSecurityProvider.  I just wanted to identify the alternative case that is being swallowed.

It is known and is still preferred to the probability of breaking an environment due to a missing property.

@RodolfoAndre  You do not need to make any additional changes for this  as far as I am concerned.

src/test/java/org/owasp/esapi/logging/appender/LogPrefixAppenderTest.java
kwwall: @RodolfoAndre - The way I handled it was to create 2 separate JUnit classes, each of which tested one of the possible values. For example, for the property **Validator.HtmlValidationAction**, which can have the value "clean" or "throw", I respectively created:

- https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/test/java/org/owasp/esapi/reference/validation/HTMLValidationRuleCleanTest.java

- https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/test/java/org/owasp/esapi/reference/validation/HTMLValidationRuleThrowsTest.java

(although for the latter, it looks as though I may have botched line 53, although I'm not sure it matters since it is set correctly in the `@Before setup` method on line 84).

But that explains the general approach of how I tested both variations. There's probably some other way to do it, but that worked. It's also why the JUnit file are named as they are.  HTH.

RodolfoAndre: Ok. So I've added the new file with the tests and reverted the other one just adding a test to check if it is working with the flag off. It is in the new commits that I've sent.

kwwall:
@RodolfoAndre - There are requested changes here. Any chance you can make them before Sunday, 2024/05/26 as I'm considering doing a release this long holiday weekend. (See Discussion #834.) If not, I will leave this PR unmerged and it will have to wait until a future release. Thanks.

kwwall:
@RodolfoAndre - If I don't hear back from you by the end of the business week (so Oct 11th, 2024), I am going to close this PR. It's been since November of last year since we've heard from you.

PR #828, opened 2024-01-23
PR to fix #824 and reference to #823

### LINKED ISSUE #824: DefaultEncoder / getCanonicalizedURI returns mix encoding for HTML special characters
### Discussed in https://github.com/ESAPI/esapi-java-legacy/discussions/823

<div type='discussions-op-text'>

<sup>Originally posted by **krog78** January 19, 2024</sup>

Hi,

DefaultEncoder / getCanonicalizedURI returns mix encoding for HTML special characters in query string (and does not seem to canonicalize the parameter value despite the fact it is mentionned):

https://github.com/ESAPI/esapi-java-legacy/blob/2136292c2225ca8ff770d85f41364873218b7ff5/src/main/java/org/owasp/esapi/reference/DefaultEncoder.java#L573

And the canonicalize is applied to scheme, host, port and also UriSegment.SCHEMSPECIFICPART, is it really relevant?

Thanks,
Regards,
Sylvain
</div>

xeno6696:
Couple of bugs discovered with DefaultEncoder.getCanonicalizedURI(URI) where 

1.)  We weren't fully handling relative URLs
2.)  A canonicalize call was occurring twice always, when logically the intent was to treat queries as their own special case of canonicalization.  Logic now splits between checking if we're dealing with a URL query and does not canonicalize twice.  

The prior bug made ESAPI prone to false positives depending on whether or not there was the presence of URL queries that were in collision with named HTMLEntities.


xeno6696:
@kwwall and @jeremiahjstacey It's my turn to wait for y'all :-D

kwwall:
@xeno6696 - I will look at this tonight, but do we have a corresponding GH issue that we can reference in this PR? If not, could you take a crack at writing up a short one. I usually reference the closed GH issues in our release notes so this will make it easier for me not to forget about this important fix. Thanks.

xeno6696:
> @xeno6696 - I will look at this tonight, but do we have a corresponding GH issue that we can reference in this PR? If not, could you take a crack at writing up a short one. I usually reference the closed GH issues in our release notes so this will make it easier for me not to forget about this important fix. Thanks.

Yeah... the issue number is in the PR title.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -985,6 +981,50 @@ public void testGetCanonicalizedUri() throws Exception {
...
+        String expectedUri = "http://palpatine@foobar.com/path_to/resource?foo=bar&para1=&amp;amp;amp;test";
+        //Please note that section 3.2.1 of RFC-3986 explicitly states not to encode
+        //password information as in http://palpatine:password@foo.com, and this will
+        //not appear in the userinfo field.
+        String input = "http://palpatine@foobar.com/path_to/resource?foo=bar&para1=&amp;amp;amp;test";
+        URI uri = new URI(input);
+        System.out.println(uri.toString());
+        assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+
+    }
```
kwwall, line 1013: Think there is a trailing tab on line 1013.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -985,6 +981,50 @@ public void testGetCanonicalizedUri() throws Exception {
...
+        //Please note that section 3.2.1 of RFC-3986 explicitly states not to encode
+        //password information as in http://palpatine:password@foo.com, and this will
+        //not appear in the userinfo field.
+        String input = "http://palpatine@foobar.com/path_to/resource?foo=bar&para1=&amp;amp;amp;test";
+        URI uri = new URI(input);
+        System.out.println(uri.toString());
+        assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+
+    }	
+	    public void testGetCanonicalizedUriWithMultQueryParams() throws Exception {
```
kwwall, line 1014: And looks like a leading tab here as well.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -985,6 +981,50 @@ public void testGetCanonicalizedUri() throws Exception {
...
+        //password information as in http://palpatine:password@foo.com, and this will
+        //not appear in the userinfo field.
+        String input = "http://palpatine@foobar.com/path_to/resource?foo=bar&para1=test";
+        URI uri = new URI(input);
+        System.out.println(uri.toString());
+        assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+
+    }	
+    
+    @org.junit.Ignore("Pre-check in unit test for issue #826")
```
kwwall, line 1000: Are you eventually planning on removing this Ignore annotation?

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -985,6 +981,50 @@ public void testGetCanonicalizedUri() throws Exception {
...
+        assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+
+    }	
+	    public void testGetCanonicalizedUriWithMultQueryParams() throws Exception {
+        System.out.println("getCanonicalizedUri");
+        Encoder e = ESAPI.encoder();
+
+        String expectedUri = "http://palpatine@foo bar.com/path_to/resource?foo=bar&bar=foo#frag";
+        //Please note that section 3.2.1 of RFC-3986 explicitly states not to encode
+        //password information as in http://palpatine:password@foo.com, and this will
```
kwwall, line 1020: Changing the 'foo.com' to 'bar.com' would make it better align with the code on line 1022. Also, I thought the RFCs were changed so that only ftp and ftps schemes supported passwords. I thought the support for passwords for http / https was dropped a while ago.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1000,6 +1040,41 @@ public void testGetCanonicalizedUriPiazza() throws Exception {
         assertEquals(expectedUri, e.getCanonicalizedURI(uri));
 
     }
+    
+    public void testIssue824() throws Exception {
+        System.out.println("getCanonicalizedUriPiazza");
+        Encoder e = ESAPI.encoder();
+
+        String expectedUri = "/webapp/ux/home?d=1705914006565&status=login&ticket=1705914090394_HzJpTROVfhW-JhRW0OqDbHu7tWXXlgrKSUmOzIMsZNCcUIiYGMXX_Q==&newsess=false&roleid=DP010101/0007&origin=ourprogram";
+        //Please note that section 3.2.1 of RFC-3986 explicitly states not to encode
```
kwwall, line 1049: Note sure why this comment is relevant to this particular test case. Am I missing something or was this just a remnant of a copy/paste error?

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1000,6 +1040,41 @@ public void testGetCanonicalizedUriPiazza() throws Exception {
...
+        //password information as in http://palpatine:password@foo.com, and this will
+        //not appear in the userinfo field.
+        String input = "/webapp/ux/home?d=1705914006565&status=login&ticket=1705914090394_HzJpTROVfhW-JhRW0OqDbHu7tWXXlgrKSUmOzIMsZNCcUIiYGMXX_Q%3D%3D&newsess=false&roleid=DP010101/0007&origin=ourprogram";
+        URI uri = new URI(input);
+        System.out.println(uri.toString());
+        assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+
+    }
+
```
kwwall, line 1058: Extraneous tab here.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1000,6 +1040,41 @@ public void testGetCanonicalizedUriPiazza() throws Exception {
...
+        //password information as in http://palpatine:password@foo.com, and this will
+        //not appear in the userinfo field.
+        String input = "/webapp/ux/home?d=1705914006565&status=login&ticket=1705914090394_HzJpTROVfhW-JhRW0OqDbHu7tWXXlgrKSUmOzIMsZNCcUIiYGMXX_Q%3D%3D&newsess=false&roleid=DP010101/0007&origin=ourprogram";
+        URI uri = new URI(input);
+        System.out.println(uri.toString());
+        assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+
+    }
+	
+    @org.junit.Ignore("Pre-check in unit test for issue #826")
```
kwwall, line 1059: See comment re: the Ignore annotation as above.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1000,6 +1040,41 @@ public void testGetCanonicalizedUriPiazza() throws Exception {
...
+	public void Issue826GetCanonicalizedDoubleAmpersand() throws Exception {
```
kwwall, line 1060: Leading tab character here.

src/test/java/org/owasp/esapi/reference/EncoderTest.java
```diff
@@ -1000,6 +1040,41 @@ public void testGetCanonicalizedUriPiazza() throws Exception {
...
+         */
+        String input = "http://127.0.0.1:3000/campaigns?goal=all&section=active&sort-by=-id&status=Draft%2C&html=&&amp;contentLaunched";
+        URI uri = new URI(input);
+        System.out.println(uri.toString());
+        try {
+        	assertEquals(expectedUri, e.getCanonicalizedURI(uri));
+        	fail();
+        } catch (Exception ex) {
+        	//Expected 
+        }
```
jeremiahjstacey, line 1076: This is a strange block.

verify the two strings are equal, then fail if they are?

would it be the same to `assertNotEquals` and remove the try/fail/catch ?

xeno6696: Really, everything in this class should be rewritten to use Expected Exceptions.  I started to do that then quickly realized that a lot of these teats are JUnit 3 and at *that* particular moment I felt like upgrading then rewriting went beyond the scope of the bug fix.  

I am guilty in this case of being lazy, I will refactor .

kwwall: Meh; I'm guilty of the same thing. If I end up editing something originally for JUnit 3, I leave it as-is. We can clean it up during the 3SAPI rewrite. Let's not make unnecessary and tedious work for ourselves.

kwwall: @jeremiahjstacey - you left a comment, but it wasn't clear if you were requesting changes or it was just a casual comment or what? If you are okay with it as it is, please login and Approve this review so I know that we're not waiting for Matt to fix something.

jeremiahjstacey: No changes required.

kwwall:
I'm going to go ahead and approve this and assume that @jeremiahjstacey is okay with that decision so we I can get it into the next release. We can clean up the test in a later follow-up, but I don't see that as a showstopper.

PR #830, opened 2024-02-04
[Snyk] Security upgrade org.owasp.antisamy:antisamy from 1.7.4 to 1.7.5

kwwall:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to fix one or more vulnerable packages in the `maven` dependencies of this project.</h3>

#### Changes included in this PR

- Changes to the following files to upgrade the vulnerable dependencies to a fixed version:
    - pom.xml

##### With an upgrade:
Severity                   | Priority Score (*)                   | Issue                   | Upgrade                   | Breaking Change                   | Exploit Maturity
:-------------------------:|-------------------------|:-------------------------|:-------------------------|:-------------------------|:-------------------------
medium severity  |  **591/1000**  <br/> **Why?** Recently disclosed, Has a fix available, CVSS 6.1  | Cross-site Scripting (XSS) <br/>[SNYK-JAVA-ORGOWASPANTISAMY-6227504](https://snyk.io/vuln/SNYK-JAVA-ORGOWASPANTISAMY-6227504) |  `org.owasp.antisamy:antisamy:` <br> `1.7.4 -> 1.7.5` <br>  |  No  | No Known Exploit 

(*) Note that the real score may have changed since the PR was raised.

Check the changes in this PR to ensure they won't cause issues with your project.

------------

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs.*

For more information:  
🧐 [View latest project report](https://app.snyk.io/org/kwwall/project/95b3b421-3d26-4a92-bbaf-c2bc1190359c?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)

🛠 [Adjust project settings](https://app.snyk.io/org/kwwall/project/95b3b421-3d26-4a92-bbaf-c2bc1190359c?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)

📚 [Read more about Snyk's upgrade and patch logic](https://support.snyk.io/hc/en-us/articles/360003891078-Snyk-patches-to-fix-vulnerabilities)

[...]

PR #833, opened 2024-02-28
fix: upgrade Antisamy to 1.7.5 to resolve CVE-2024-23635

mpreziuso:
This is to resolve CVE-2024-23635 relating to Antisamy 1.7.4.
It's related and similar to https://github.com/ESAPI/esapi-java-legacy/pull/830, however it resolves dependency conflicts and updates 2 test cases as the output we get from Antisamy looks a bit different. I think this is to be expected: as mentioned in the [release notes](https://github.com/nahsra/antisamy/releases/tag/v1.7.5) for Antisamy 1.7.5 the output may have changed again:
> Note: The upgrade in the HTML parser may alter outputs compared to 1.7.4 and before. This may impact regression tests that involve AntiSamy if they are too strict when comparing a resulting output with the expected one.

Also related to: https://github.com/nahsra/antisamy/issues/389 and https://github.com/nahsra/antisamy/pull/388.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -243,12 +243,17 @@
         <dependency>
             <groupId>org.owasp.antisamy</groupId>
             <artifactId>antisamy</artifactId>
-            <version>1.7.4</version>
+            <version>1.7.5</version>
             <exclusions>
                 <!-- excluded because we directly import newer version below. -->
                 <exclusion>
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-api</artifactId>
+                </exclusion>
+                <!-- excluded because commons-beanutils imports a newer version. -->
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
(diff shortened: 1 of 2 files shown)


kwwall (approved):
LGTM. Thanks for doing this.

kwwall:
@mpreziuso - Because there are no actual code changes here that will go into the ESAPI jar (the few code changes were simply adjustments to JUnit tests caused by the update to AntiSamy 1.7.5), I am not going to ask you to go back and do this PR over and sign your commits, but for a while now, we have been requiring signed commits. I would suggest reading through https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification and then creating and uploading your public signing key. Had this been code changes that would have ended up in the ESAPI jar though, I would have rejected this PR until this was corrected.

kwwall:
An important note: **CVE-2024-23635** does **_NOT_** impact ESAPI unless you have  added AntiSamy's '`preserveComments`' directive to ESAPI's AntiSamy policy file, "**antisamy-esapi.xml**", which I would consider similar to intentionally sabotaging yourself. For further details, please see my post "[Before you panic - New AntiSamy release available](https://groups.google.com/a/owasp.org/g/esapi-project-users/c/vH7kM4mUTGk)" made to the _esapi-project-users_ Google group.

PR #840, opened 2024-04-26
Issue #839 JavaLogFactory ConcMod

### LINKED ISSUE #839: ConcurrentModificationException
#### Describe the bug
JavaLogFactory.java's readLoggerConfiguration function throws ConcurrentModificationException at this line 
```java
if (System.getProperties().keySet().stream().anyMatch(propKey ->
        "java.util.logging.config.class".equals(propKey) || "java.util.logging.config.file".equals(propKey)))
```
in a multi-threaded environment when another thread is concurrently setting system properties while ESAPI is iterating over system properties.

2.5.2.0

#### To Reproduce
```java
public class Reproduce {
    public static void main(String[] args) {
        Thread t = new Thread(new Runnable() {
            public void run() {
                while (true) {
                    long time = System.currentTimeMillis();
                    // Emulate another thread modifying system properties
                    System.setProperty(time + "", time + "");
                    try {
                        Thread.sleep(500);
                    } catch (InterruptedException e) {
                        e.printStackTrace();
                    }
                }
            }
        }); t.start();
        EsapiThread esapiThread = new EsapiThread();
        esapiThread.start();
    }

    private static class EsapiThread extends Thread {
        public void run() {
            // Emulate the code of JavaLogFactory in ESAPI
            System.getProperties().keySet().stream().anyMatch(a -> {
                try {
                    Thread.sleep(2000);
                } catch (InterruptedException e) {
                    throw new RuntimeException(e);
                }
                return "dummy".equals(a);
            });
        }
    }
}

```

#### Expected behavior
Proposed change: copy the `System.getProperties().keySet()` to a new collection and iterating over the new collection instead of the original one to avoid thread safety problem.

#### Platform environment (please complete the following information)
 - JDK version: JDK8

jeremiahjstacey:
Removing support for esapi-java-logging.properties file from baseline.

ConfigurationException is thrown if file is found on the path at runtime.  Exception message links to a wiki page with instructions on how to configure the application instance.

Closes #839 
Also note the wiki page  has been created.
https://github.com/ESAPI/esapi-java-legacy/wiki/Configuring-the-JavaLogFactory

### DIFF
--- a/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
+++ b/src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
@@ -14,8 +14,16 @@
  */
 package org.owasp.esapi.logging.java;
 
+import static org.owasp.esapi.PropNames.APPLICATION_NAME;
+import static org.owasp.esapi.PropNames.LOG_APPLICATION_NAME;
+import static org.owasp.esapi.PropNames.LOG_CLIENT_INFO;
+import static org.owasp.esapi.PropNames.LOG_ENCODING_REQUIRED;
+import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
+import static org.owasp.esapi.PropNames.LOG_USER_INFO;
+
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -25,7 +33,6 @@ import java.util.logging.LogManager;
 import org.owasp.esapi.ESAPI;
 import org.owasp.esapi.LogFactory;
 import org.owasp.esapi.Logger;
-import org.owasp.esapi.PropNames;
 import org.owasp.esapi.codecs.HTMLEntityCodec;
 import org.owasp.esapi.errors.ConfigurationException;
 import org.owasp.esapi.logging.appender.LogAppender;
@@ -34,13 +41,6 @@ import org.owasp.esapi.logging.cleaning.CodecLogScrubber;
 import org.owasp.esapi.logging.cleaning.CompositeLogScrubber;
 import org.owasp.esapi.logging.cleaning.LogScrubber;
 import org.owasp.esapi.logging.cleaning.NewlineLogScrubber;
-
-import static org.owasp.esapi.PropNames.LOG_ENCODING_REQUIRED;
-import static org.owasp.esapi.PropNames.LOG_USER_INFO;
-import static org.owasp.esapi.PropNames.LOG_CLIENT_INFO;
-import static org.owasp.esapi.PropNames.LOG_APPLICATION_NAME;
-import static org.owasp.esapi.PropNames.APPLICATION_NAME;
-import static org.owasp.esapi.PropNames.LOG_SERVER_IP;
 
 /**
  * LogFactory implementation which creates JAVA supporting Loggers.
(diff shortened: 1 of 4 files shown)


kwwall:
@jeremiahjstacey - Do you think it's useful to mention the previous defaults that were in the (now removed -- after this PR is merged) esapi-java-logging.properties file that used to be in the ESAPI configuration jar? I thought we may get questions about what properties used to be there for JUL and thought we could drop those into the new wiki page (https://github.com/ESAPI/esapi-java-legacy/wiki/Configuring-the-JavaLogFactory) that you wrote.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -93,43 +93,24 @@ public class JavaLogFactory implements LogFactory {
...
+         * 
+         * Verify the file is not present and fail if found to enforce understanding of
+         * the configuration method.
          */
         try (InputStream stream = JavaLogFactory.class.getClassLoader().
                 getResourceAsStream("esapi-java-logging.properties")) {
-            if (stream == null) {
-                throw new ConfigurationException("Unable to locate resource: esapi-java-logging.properties");
+            if (stream != null) {
+                throw new ConfigurationException("esapi-java-logging.properties is no longer supported.  See https://github.com/ESAPI/esapi-java-legacy/wiki/Configuring-the-JavaLogFactory for information on corrective actions.");
```
kwwall, line 106: I think it would make more sense to define this exception message as `final String ` and the reference it both places. That way, if we every decide to change the error message, we won't forget to change one of the instances. Otherwise, I think this is great and hopefully it will head off questions from those who seem to have a habit of never reading the release notes.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -93,43 +93,24 @@ public class JavaLogFactory implements LogFactory {
...
             }
-            logManager.readConfiguration(stream);
+
         } catch (IOException ioe) {
-            throw new ConfigurationException("Failed to load esapi-java-logging.properties.", ioe);
+            // This is a little strange, I know.
+            // If the IOException is thrown, then the file actually exists but is malformatted or has some other issue.
+            // The file should not exist at all, so use the same message as above but include the original exception in the log as well.
```
kwwall, line 112: See previous comment so only have to change it in one place to ensure consistency.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
```diff
@@ -48,83 +43,6 @@ public class JavaLogFactoryTest {
     @Rule
     public ExpectedException exEx = ExpectedException.none();
 
-    @Test
```
kwwall, line 51: So, we're not going to try to test the case if the `ConfigurationException` is thrown if the file **esapi-java-logging.properties** is found. As long as you tested all the cases separately, I guess I'm okay with that, otherwise the testing gets messy because you have to create the file, test, delete it, and test again.

kwwall (approved):
@jeremiahjstacey - Some minor suggestions is all, but I'm okay with this PR as it is.

@xeno6696 - I may try doing a release over Memorial Day weekend (no promise though), so if I will give you until Sat around noon if you want to review this. Ideally, just leave an 'approved' as early as possible so I know you're good with it.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
jeremiahjstacey: Yeah, I went light on the testing with this change as we'd have to be testing a negative condition.

I am relying on the implied tested state from all of the other tests - knowing that the properties file has been removed and the configuration has changed, all existing tests continue to function.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: Update applied

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
jeremiahjstacey: Tracking in conversation above.  Resolving this thread in favor of the previous conversation.

src/test/java/org/owasp/esapi/logging/java/JavaLogFactoryTest.java
xeno6696: Sorry I'm late to the party.  Kids birthday this weekend and family's in town.  

Regression tests FTW.  I have no issues here.

xeno6696 (approved):

Issue #844, opened 2024-06-11
Update the logging properties to opt-out of the prefix events


mickeyz07:
**I would like to disable the usage of [SECURITY SUCCESS -> /com.microservices.project.handlers.SecurityHandler]**

Before the usage of ESAPI the logged line was:
`2024-05-03 15:14:13,912 INFO c.t.m.e.h.SecurityHandler [pool-3-thread-1] Successfully transformed security update #1`

With the ESAPI the logged line becomes:
`2024-05-03 15:14:13,912 INFO [pool-3-thread-1] o.o.e.l.s.Slf4JLogLevelHandlers$3: [SECURITY SUCCESS -> /com.microservices.project.handlers.SecurityHandler] Successfully transformed security update #1`

**As Jeremiah J. Stacey confirmed in our email conversation, following needs to be done:**

- The system default should maintain the current capability if a property is not provided.
- The property should be read in as part of the LogFactory initialization process, and passed to the constructor of the LogPrefixAppender
  - This will need to be applied to all of the LogFactory implementations for consistency.
- The LogPrefixAppender will need to have a new argument to its constructor, and a new member variable too store the reference
- The LogPrefixAppender.appendTo method will need to be updated to set the boolean configurations on both the EventTypeLogSupplier and the ServerInfoSupplier
- The EventTypeLogSupplier should be expanded to provide a public setter for the boolean configuration on whether to log, and the get method will need to be updated to return an empty string when the event type is disabled.
- The ServerInfoSupplier will need an additional public setter for the classpath optional set, and the get method will need to be updated for that additional logical path.
- Tests will need to be added for the new APIs for each of the suppliers to show the behavior when both enabled and disabled match expectations.

kwwall:
You do realize that this will also affect ESAPI's internal logging,
correct? Doing so may effect the alerting on any SIEM you may be usually.
Just wanted you to beware.
-kevin

kwwall:
Closed via PR #845

PR #845, opened 2024-06-25
Update the logging properties to opt-out of the prefix events #844

mickeyz07:
Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
If all Logger.* settings in the ESAPI.properties file are false (including Logger.LogPrefix), then the log output would be the same as if no ESAPI was used.
That can be important, especially for clients that are using cloud providers. The charges for extra logging can quickly add up.
So, the user can sanitise untrusted data used to construct log entries by using a safe logging mechanism in the OWASP ESAPI Logger without any extra logging information.
However, only by changing a couple of settings in the ESAPI.properties file, a user can have a very detailed log if required.

### DIFF
--- a/configuration/esapi/ESAPI.properties
+++ b/configuration/esapi/ESAPI.properties
@@ -408,8 +408,8 @@ Logger.UserInfo=true
 Logger.ClientInfo=true
 
 # Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
-# If all above Logger entries are set to false and LogIgnorePrefix is true, then the output would be the same like if no ESAPI was used
-Logger.LogIgnorePrefix=true
+# If all above Logger entries are set to false, as well as LogPrefix, then the output would be the same as if no ESAPI was used
+Logger.LogPrefix=true
 
 #===========================================================================
 # ESAPI Intrusion Detection
--- a/configuration/esapi/ESAPI.properties
+++ b/configuration/esapi/ESAPI.properties
@@ -407,6 +407,10 @@ Logger.UserInfo=true
 # Determines whether ESAPI should log the session id and client IP.
 Logger.ClientInfo=true
 
+# Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
+# If all above Logger entries are set to false and LogIgnorePrefix is true, then the output would be the same like if no ESAPI was used
+Logger.LogIgnorePrefix=true
+
 #===========================================================================
 # ESAPI Intrusion Detection
 #
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoader.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoader.java
@@ -34,6 +34,13 @@ public interface EsapiPropertyLoader {
     public Boolean getBooleanProp(String propertyName) throws ConfigurationException;
 
     /**
+     * Get any Boolean type property from security configuration.
+     * If property does not exist in configuration or has incorrect type, defaultValue is returned
+     * @return property value.
+     */
+    public Boolean getBooleanProp(String propertyName, Boolean defaultValue);
+
+    /**
      * Get any property from security configuration. As every property can be returned as string, this method
      * throws exception only when property does not exist.
      *
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoader.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoader.java
@@ -34,13 +34,6 @@ public interface EsapiPropertyLoader {
     public Boolean getBooleanProp(String propertyName) throws ConfigurationException;
 
     /**
-     * Get any Boolean type property from security configuration.
-     * If property does not exist in configuration or has incorrect type, defaultValue is returned
-     * @return property value.
-     */
-    public Boolean getBooleanProp(String propertyName, Boolean defaultValue);
-
-    /**
      * Get any property from security configuration. As every property can be returned as string, this method
      * throws exception only when property does not exist.
      *
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
@@ -80,18 +80,6 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
      * {@inheritDoc}
      */
     @Override
-    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
-        for (AbstractPrioritizedPropertyLoader loader : loaders) {
-            return loader.getBooleanProp(propertyName, defaultValue);
-        }
-        return defaultValue;
-    }
-
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
     public String getStringProp(String propertyName) throws ConfigurationException {
         for (AbstractPrioritizedPropertyLoader loader : loaders) {
             try {
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
@@ -80,6 +80,22 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
      * {@inheritDoc}
      */
     @Override
+    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
+        for (AbstractPrioritizedPropertyLoader loader : loaders) {
+            try {
+                return loader.getBooleanProp(propertyName);
+            } catch (ConfigurationException e) {
+                return defaultValue;
+            }
+        }
+        return defaultValue;
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
     public String getStringProp(String propertyName) throws ConfigurationException {
         for (AbstractPrioritizedPropertyLoader loader : loaders) {
             try {
--- a/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
+++ b/src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
@@ -82,11 +82,7 @@ public class EsapiPropertyManager implements EsapiPropertyLoader {
     @Override
     public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
         for (AbstractPrioritizedPropertyLoader loader : loaders) {
-            try {
-                return loader.getBooleanProp(propertyName);
-            } catch (ConfigurationException e) {
-                return defaultValue;
-            }
+            return loader.getBooleanProp(propertyName, defaultValue);
         }
         return defaultValue;
     }
--- a/src/main/java/org/owasp/esapi/configuration/StandardEsapiPropertyLoader.java
+++ b/src/main/java/org/owasp/esapi/configuration/StandardEsapiPropertyLoader.java
@@ -74,24 +74,6 @@ public class StandardEsapiPropertyLoader extends AbstractPrioritizedPropertyLoad
      * {@inheritDoc}
      */
     @Override
-    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
-        String property = properties.getProperty(propertyName);
-        if (property == null) {
-            return defaultValue;
-        }
-        if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
-            return true;
-        }
-        if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
-            return false;
-        }
-        return defaultValue;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
     public String getStringProp(String propertyName) throws ConfigurationException {
         String property = properties.getProperty(propertyName);
         if (property == null) {
--- a/src/main/java/org/owasp/esapi/configuration/StandardEsapiPropertyLoader.java
+++ b/src/main/java/org/owasp/esapi/configuration/StandardEsapiPropertyLoader.java
@@ -74,6 +74,24 @@ public class StandardEsapiPropertyLoader extends AbstractPrioritizedPropertyLoad
      * {@inheritDoc}
      */
     @Override
+    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
+        String property = properties.getProperty(propertyName);
+        if (property == null) {
+            return defaultValue;
+        }
+        if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
+            return true;
+        }
+        if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
+            return false;
+        }
+        return defaultValue;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
     public String getStringProp(String propertyName) throws ConfigurationException {
         String property = properties.getProperty(propertyName);
         if (property == null) {
--- a/src/main/java/org/owasp/esapi/configuration/XmlEsapiPropertyLoader.java
+++ b/src/main/java/org/owasp/esapi/configuration/XmlEsapiPropertyLoader.java
@@ -90,24 +90,6 @@ public class XmlEsapiPropertyLoader extends AbstractPrioritizedPropertyLoader {
      * {@inheritDoc}
      */
     @Override
-    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
-        String property = properties.getProperty(propertyName);
-        if (property == null) {
-            return defaultValue;
-        }
-        if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
-            return true;
-        }
-        if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
-            return false;
-        }
-        return defaultValue;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
     public String getStringProp(String propertyName) throws ConfigurationException {
         String property = properties.getProperty(propertyName);
         if (property == null) {
--- a/src/main/java/org/owasp/esapi/configuration/XmlEsapiPropertyLoader.java
+++ b/src/main/java/org/owasp/esapi/configuration/XmlEsapiPropertyLoader.java
@@ -90,6 +90,24 @@ public class XmlEsapiPropertyLoader extends AbstractPrioritizedPropertyLoader {
      * {@inheritDoc}
      */
     @Override
+    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
+        String property = properties.getProperty(propertyName);
+        if (property == null) {
+            return defaultValue;
+        }
+        if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
+            return true;
+        }
+        if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
+            return false;
+        }
+        return defaultValue;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
     public String getStringProp(String propertyName) throws ConfigurationException {
         String property = properties.getProperty(propertyName);
         if (property == null) {
--- a/src/main/java/org/owasp/esapi/logging/appender/EventTypeLogSupplier.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/EventTypeLogSupplier.java
@@ -31,7 +31,7 @@ public class EventTypeLogSupplier // implements Supplier<String>
     /** EventType reference to supply log representation of. */
     private final EventType eventType;
     /** Whether to log or not the event type */
-    private boolean ignoreLogEventType = false;
+    private boolean logEventType = true;
 
     /**
      * Ctr
@@ -44,13 +44,10 @@ public class EventTypeLogSupplier // implements Supplier<String>
 
     // @Override    -- Uncomment when we switch to Java 8 as minimal baseline.
     public String get() {
-        if (this.ignoreLogEventType) {
-            return "";
-        }
-        return eventType.toString();
+        return logEventType ? eventType.toString() : "";
     }
 
-    public void setIgnoreLogEventType(boolean ignoreLogEventType) {
-        this.ignoreLogEventType = ignoreLogEventType;
+    public void setLogEventType(boolean logEventType) {
+        this.logEventType = logEventType;
     }
 }
--- a/src/main/java/org/owasp/esapi/logging/appender/EventTypeLogSupplier.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/EventTypeLogSupplier.java
@@ -30,18 +30,27 @@ public class EventTypeLogSupplier // implements Supplier<String>
 {
     /** EventType reference to supply log representation of. */
     private final EventType eventType;
+    /** Whether to log or not the event type */
+    private boolean ignoreLogEventType = false;
 
     /**
      * Ctr
      *
-     * @param evtyp EventType reference to supply log representation for
+     * @param eventType EventType reference to supply log representation for
      */
-    public EventTypeLogSupplier(EventType evtyp) {
-        this.eventType = evtyp == null ? Logger.EVENT_UNSPECIFIED : evtyp;
+    public EventTypeLogSupplier(EventType eventType) {
+        this.eventType = eventType == null ? Logger.EVENT_UNSPECIFIED : eventType;
     }
 
     // @Override    -- Uncomment when we switch to Java 8 as minimal baseline.
     public String get() {
+        if (this.ignoreLogEventType) {
+            return "";
+        }
         return eventType.toString();
+    }
+
+    public void setIgnoreLogEventType(boolean ignoreLogEventType) {
+        this.ignoreLogEventType = ignoreLogEventType;
     }
 }
--- a/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
@@ -46,14 +46,11 @@ public class LogPrefixAppender implements LogAppender {
      * @param logServerIp        Whether or not to record server ip information
      * @param logApplicationName Whether or not to record application name
      * @param appName            Application Name to record.
+     * @param logPrefix          is set by default to true
      */
+    @SuppressWarnings("JavadocReference")
     public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName) {
-        this.logUserInfo = logUserInfo;
-        this.logClientInfo = logClientInfo;
-        this.logServerIp = logServerIp;
-        this.logApplicationName = logApplicationName;
-        this.appName = appName;
-        this.logPrefix = true;
+        this(logUserInfo, logClientInfo, logServerIp, logApplicationName, appName, true);
     }
 
     /**
--- a/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
+++ b/src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
@@ -39,7 +39,7 @@ public class LogPrefixAppender implements LogAppender {
     private final boolean logPrefix;
 
     /**
-     * Ctr.
+     * Constructor
      *
      * @param logUserInfo      Whether or not to record user information
      * @param logClientInfo      Whether or not to record client information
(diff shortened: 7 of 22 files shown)


configuration/esapi/ESAPI.properties
```diff
@@ -407,6 +407,10 @@ Logger.UserInfo=true
 # Determines whether ESAPI should log the session id and client IP.
 Logger.ClientInfo=true
 
+# Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
+# If all above Logger entries are set to false and LogIgnorePrefix is true, then the output would be the same like if no ESAPI was used
+Logger.LogIgnorePrefix=true
```
jeremiahjstacey, line 412: The boolean logic / variable name is a little confusing.

I think the logic will be easier to manage if we phrase this as a Positive statement:
```
# Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
Logger.LogPrefix=true  
```

I'd prefer to remove the additional comment from line 411, as that may lead to confusion around the "other log entries" if (and when) the properties are expanded.

Changing the naming to a positive statement will also apply throughout the naming of the implementation, and the  boolean logic checks.

src/main/java/org/owasp/esapi/logging/appender/ServerInfoSupplier.java
```diff
@@ -57,10 +58,19 @@ public String get() {
                 appInfo.append(request.getLocalAddr()).append(":").append(request.getLocalPort());
             }
         }
-        if (logAppName) {
-            appInfo.append("/").append(applicationName);
+
+        if (this.logAppName) {
+            if (this.applicationName != null && !this.applicationName.isEmpty()) {
```
kwwall, line 63: @jeremiahjstacey and @xeno6696 - Seeing this common boilerplate reminds me to let's make a note to use Lombok annotations in ESAPI 3.x. I'd do it now, but I really don't want to add another dependency at this point until we can break things up into smaller components and reduce the total dependencies where you are just using small pieces.

src/test/java/org/owasp/esapi/logging/appender/EventTypeLogSupplierIgnoreEventTypeTest.java
```diff
@@ -0,0 +1,44 @@
...
+        return paramSets;
+    }
+
+    private final Logger.EventType eventType;
+    private final String expectedResult;
+
+    public EventTypeLogSupplierIgnoreEventTypeTest(Logger.EventType eventType, String result) {
+        this.eventType = eventType;
+        this.expectedResult = result;
+    }
```
kwwall, line 37: Suggestion: Add blank line here for improved readability.

src/test/resources/esapi/ESAPI.properties
```diff
@@ -439,6 +439,10 @@ Logger.UserInfo=true
 # Determines whether ESAPI should log the session id and client IP.
 Logger.ClientInfo=true
 
+# Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
+# If all above Logger entries are set to false and LogIgnorePrefix is true, then the output would be the same like if no ESAPI was used
+Logger.LogIgnorePrefix=true
```
kwwall, line 444: Change this property name in accordance with @jeremiahjstacey's suggestion in c**onfiguration/esapi/ESAPI.properties.**

kwwall (commented):
So far, no objections, but I need to think about the tests a bit more. In the meantime, @mickeyz07, perhaps you can respond to my comments.

src/main/java/org/owasp/esapi/logging/appender/EventTypeLogSupplier.java
```diff
@@ -30,18 +30,27 @@ public class EventTypeLogSupplier // implements Supplier<String>
...
+        this.eventType = eventType == null ? Logger.EVENT_UNSPECIFIED : eventType;
     }
 
     // @Override    -- Uncomment when we switch to Java 8 as minimal baseline.
     public String get() {
+        if (this.ignoreLogEventType) {
+            return "";
+        }
         return eventType.toString();
     }
```
jeremiahjstacey, line 51: Preferred to have single returns from methods. 
Ternary recommended (please note that my sample is applying the updated property name):
`return logEventType ? eventType.toString() ? "";`

src/main/java/org/owasp/esapi/logging/appender/LogPrefixAppender.java
```diff
@@ -51,11 +53,32 @@ public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean log
         this.logServerIp = logServerIp;
         this.logApplicationName = logApplicationName;
         this.appName = appName;
+        this.logPrefix = true;
```
jeremiahjstacey, line 56: I'd be in favor of removing this constructor entirely. That can have impacts if anyone has actually extended the class, so I suspect that we would not want to do that.

Assuming that we cannot delete this code, I'd like to request two more adjustments here:
1.  Update the Javadoc to specify the logPrefix value is configured to true
2. Chain the constructors - have this one call the other
```
public LogPrefixAppender(boolean logUserInfo, boolean logClientInfo, boolean logServerIp, boolean logApplicationName, String appName) {
  this (logUserInfo, logClientInfo, logServerIp, logApplicationName, appName, true);
}
```

src/main/java/org/owasp/esapi/logging/appender/ServerInfoSupplier.java
```diff
@@ -57,10 +58,19 @@ public String get() {
...
+                appInfo.append("/").append(this.applicationName);
+            }
+            else if (this.applicationName == null) {
+                appInfo.append("/").append(this.applicationName);
+            }
```
jeremiahjstacey, line 68: this feels incorrect.
If the applicationname is null then we want to log it?

src/test/java/org/owasp/esapi/logging/appender/LogPrefixAppenderTest.java
```diff
@@ -34,6 +34,7 @@ public class LogPrefixAppenderTest {
     private String testLogMessage =  testName.getMethodName() + "-MESSAGE";
     private String testApplicationName = testName.getMethodName() + "-APPLICATION_NAME";
     private EventType testEventType = Logger.EVENT_UNSPECIFIED;
+    //private boolean testIgnorePrefix = true;
```
jeremiahjstacey, line 37: clean up commented line

src/main/java/org/owasp/esapi/logging/appender/ServerInfoSupplier.java
jeremiahjstacey: Looking closer at this,  I'm not sure the inner if/elsif block adds value?

//If not null and not empty -> Log the app name
// else if null -> Log the app name

I think it makes sense to remove the  applicationName validation and  log the applicationName if it's configured to log.

```
if (this.logAppName) {
  appInfo.append("/").append(this.applicationName);
}
```

If the log does not appear correctly, that is a different configuration that should be made in the environment.

kwwall: If `applicationName` is `null` or the empty string, perhaps we should log it as "Unknown"?

jeremiahjstacey: If that were the case, then we'd want to change the default value returned from the SecurityConfiguration so that an unset ApplicationName would be consistent throughout the instance.
I think that's out of scope for this work, and I think we should log the value that was pulled from the config (even if that is null). 
Note that the ApplicationName is pulled from the SecurityConfiguration in each of the LogFactory implementations:
` String appName = ESAPI.securityConfiguration().getStringProp(APPLICATION_NAME);`

jeremiahjstacey:
@kwwall - This all looks good to me.  if your concerns have been addressed please merge or leave a comment and I will follow up and get it merged this week

src/main/java/org/owasp/esapi/logging/appender/ServerInfoSupplier.java
kwwall: @jeremiahjstacey - Fair enough. But maybe something we should keep in mind for the future.

kwwall:
> @kwwall - This all looks good to me. if your concerns have been addressed please merge or leave a comment and I will follow up and get it merged this week

@jeremiahjstacey - I will take care of merging this, but probably won't get to it by today though, as I had hoped. But I want to run a few dynamic tests first to make sure the defaults are all set so as to be backward compatible. I think they are, but it's a lot easier to try it than try to judge based on the code. I hope to have it merged by this weekend unless I find something that needs fixed

kwwall:
@mickeyz07 - Sorry I didn't notice this before. Jeremiah and I were focused on the code review itself. However, I just noticed that you have not signed your commits with a GPG / PGP key. For commits that will be merged to the 'develop' or 'main' branches, we require those commits to be signed.

From our **CONTRIBUTING-TO-ESAPI.txt** file:
```text
A Special Note Regarding Making Commits for PRs
    Shortly after the 2.5.1.0 ESAPI release in late November 2022, the ESAPI
    team decided to lock down the 'develop' amd 'main' branches. Merges from
    PRs are done to the 'develop' branch. That means that if you intend to
    contribute to ESAPI, you must be signing your commits. Please see the
    GitHub instructions at
        https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
    for details.
```

There are detailed instructions on how to set this up in our [ESAPI Release Steps document.](https://github.com/ESAPI/esapi-java-legacy/blob/develop/documentation/ESAPI-release-steps.pdf) Unfortunately, I seem to recall this before. GitHub doesn't warn us about this until after we have 2 people from the team approve the PR. I think it would then probably then force you to go back and set up signing and you'd have to made at least one commit to all that you've modified and then we'd have to get 2 of us to approve it again since there would be a commit after the approval.

So, if you could, could you take care of that. I'm not sure if it would fix it though, because I think ti may require ALL commits to be signed, not just the last one in a PR. (@xeno6696 or @jeremiahjstacey - do either of you know?)

I'm guessing it may not have flagged this because you forked it and created it on your 'issue-844' branch (which, was as instructed). I think there is an option where I can set up our repo to require signed commits on ALL branches, but I didn't do that because I figured it my break things like Dependabot and Snyk generated PRs.

Anyhow, can you see if you can somehow fix this? Thanks.

xeno6696:
There's no reason we couldn't have a "signature" commit, just a simple merge with a text change detailing it as a signature.  Just to get it into history and I think that ought to be sufficient.  Going back and resigning old commits is more trouble than its worth.
________________________________

kwwall:
@xeno6696 - That's sort of what I was thinking of having him do, but I wasn't sure if that would pass the muster of the GitHub signature verification. I thought it required every single commit to be signed if it was to end up on the 'develop' or 'main' branch. That said, maybe it would work if we did a squash-merge? I don't know.

kwwall:
So, @mickeyz07 found a severe **must fix** issue in your code. I made this simple change to simulate users NOT updating their ESAPI.properties file (which I can guarantee you that many will do):
```bash
$ git diff src/test/resources/esapi/ESAPI.properties
diff --git a/src/test/resources/esapi/ESAPI.properties b/src/test/resources/esapi/ESAPI.properties
index 8ffc61f6..a84328ab 100644
--- a/src/test/resources/esapi/ESAPI.properties
+++ b/src/test/resources/esapi/ESAPI.properties
@@ -441,7 +441,7 @@ Logger.ClientInfo=true
 
 # Determines whether ESAPI should log the prefix of [EVENT_TYPE - APPLICATION NAME].
 # If all above Logger entries are set to false, as well as LogPrefix, then the output would be the same as if no ESAPI was used
-Logger.LogPrefix=true
+# Logger.LogPrefix=true
 
 #===========================================================================
 # ESAPI Intrusion Detection
```
And when I re-ran 'mvn test' I got this:
```text
[ERROR] Tests run: 4256, Failures: 7, Errors: 2434, Skipped: 0
```
A similar error happens if one sets the value of `Logger.LogPrefix` to anything other than `true` or `false`. They are all throwing ConfigurationExceptions some of which end up as `ExceptionInInitializerError`. That's why I said it was important that regardless of the value, I want to treat the absence of this property value or any invalid property value as it was in 2.5.4.0. So, it they put in:
```bash
Logger.LogPrefix=off
```
or even something absurd as:
```bash
Logger.LogPrefix=scoopy_do
```
you have to treat is as though they had:
```bash
Logger.LogPrefix=true
```
It's okay if you want to output an error to stdout using `logSpecial` or whatever it is that we normally use for that, but we can't be throwing exceptions that will kill their application just because they haven't followed instructions. Because many won't.

Make sense?

kwwall (changes requested):
Changes discussed in https://github.com/ESAPI/esapi-java-legacy/pull/845#issuecomment-2211793935 are required before this will be approved.

mickeyz07:
@kwwall @jeremiahjstacey I have made changes and signed the commit with GPG key.
Currently if any of the Logger.* properties is missing or it has value different then true, false, yes, no the tests are failing. 
So to reiterate - the problem is not only with the Logger.LogPrefix.
As you could see in the code I have made provision for all Logger.* properties. 
Only Logger.LogEncodingRequired=false.
All other Boolean Logger.* properties have the default value of true.
Not sure if you require to extend that behavior to other properties? 
So if the value for Boolean properties is not valid, instead of throwing an exception, now reverts the value for the property to the default value?
In that case we could have another function with additional default parameter:
public Boolean getBooleanProp(String propertyName, boolean value)
That  would require additional testing.
So for this PR, the changes that I made are taking care of the Logger.* properties.
All tests that Kevin "broke" with commenting out Logger.* are now working correctly.

src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
```diff
@@ -1441,8 +1441,17 @@ public Boolean getBooleanProp(String propertyName) throws ConfigurationException
         try {
             return esapiPropertyManager.getBooleanProp(propertyName);
         } catch (ConfigurationException ex) {
+
             String property = properties.getProperty( propertyName );
             if ( property == null ) {
+                if (propertyName.startsWith("Logger.")) {
```
kwwall, line 1447: @mickeyz07 -- IMO, this is a major implementation kludge to me. Generally, I was expecting in whatever static initializer that checks the value for the `Logger.LogPrefix` property, that any error / exception that would be thrown that you would just treat it as 'true' and then log something via `logSpecial`. If for some reason that approach turns out to be impossible, then the next best alternative would be to create a special dedicated getLogPrefix method. Another preferred alternative (something I've been meaning to do for a while now) is to create a new `Boolean getBooleanProp(String propName, boolean default)` and use that, and just pass in 'true' for the default, so if the specified property name is not a boolean, it defaults to the specified default value (which should be true for `Logger.LogPrefix.` I am also very confused why this is even written the way that it it; it appears that you could omit the entire:
```java
                if (propertyName.startsWith("Logger.")) {
```
and just went with the specific property name check that follows. (And why _this_ specific check? I would have expected `Logger.LogPrefix` here, since that's the new property. So why `Logger.LogEncodingRequired`? I'm confused.)

@jeremiahjstacey - I'd like your opinion on this, but I don't like this **_at all._** Can your think of any alternatives.

kwwall (commented):
Not going to approve this unless Matt or Jeremiah find this kludge acceptable.

src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
jeremiahjstacey: I think I agree @kwwall  - Updates in this file will lead to additional scope that I think we'll be better off not trying to tackle in this effort.

@mickeyz07 - My understanding of what the desire here is to try to solve the problem at the lowest level possible.  I think I agree with the intent, but I think you can agree that it's not the cleanest or most maintainable due to current restrictions in the library's interfaces.
To facilitate a faster integration of this feedback, I would like to propose the following:
1.  Revert all changes to DefaultSecurityConfiguration.java
2. Update JavaLogFactory and SLF4JLogFactory implementations to default to TRUE, attempt to read from properties, and useLogSpecial on failure

```
boolean logPrefix = true;
try {
  logPrefix = ESAPI.securityConfiguration().getBooleanProp(LOG_PREFIX);
} catch (ConfigurationException ex) {
ESAPI.securityConfiguration().logSpecial("Failed to read Log Prefix configuration. Defaulting to enabled. ", ex);
}
```

It's not the best possible solution, but I think that's the best we can do in the current library implementation.

kwwall: Okay, my thoughts, exactly, and what you described was my preferred approach at resolving this.

@mickeyz07 - Please make the change that @jeremiahjstacey detailed above. Thanks.

kwwall (changes requested):
Changes requested. See previous comments by Jeremiah and myself. It should be pretty straightforward since he pretty much laid out the code changes.

mickeyz07:
@kwwall @jeremiahjstacey I have added the new method getBooleanProp(String propertyName, boolean value)
I have used that method only for Logger.LogPrefix, e.g.
        boolean logServerIp = ESAPI.securityConfiguration().getBooleanProp(LOG_SERVER_IP);
        boolean logPrefix = ESAPI.securityConfiguration().getBooleanProp(LOG_PREFIX, true);
All the tests run successfully.

kwwall:
@mickeyz07 We are getting close, so please don't give up. But there are 2 things that I see missing right out of the gate:

1. You need to add the new `getBooleanProp` to src/test/java/org/owasp/esapi/SecurityConfigurationWrapper.java
2. You need to add some JUnit tests for the new method. Probably put those tests in the following files where other tests for the original `getBooleanProp` method resides:
    * src/test/java/org/owasp/esapi/configuration/EsapiPropertyManagerTest.java
    * src/test/java/org/owasp/esapi/configuration/StandardEsapiPropertyLoaderTest.java
    * src/test/java/org/owasp/esapi/configuration/XmlEsapiPropertyLoaderTest.java

mickeyz07:
@kwwall I have added the required tests for all 3 configuration test files you noted. Please check.

src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoader.java
```diff
@@ -33,6 +33,13 @@ public interface EsapiPropertyLoader {
      */
     public Boolean getBooleanProp(String propertyName) throws ConfigurationException;
 
+    /**
+     * Get any Boolean type property from security configuration.
+     * If property does not exist in configuration or has incorrect type, defaultValue is returned
```
kwwall, line 38: From the Javadoc: for this interface: "or has incorrect type". 

Well, that is the $64k question, isn't it? What is the _INTUITIVE_ behavior if one calls something like:
```java
    Boolean propValue = getBoolean("SomePropName", true)
```
but somewhere, say in **ESAPI.properties** or elsewhere, we have defined
```bash
SomePropName=This is not a boolean property
```
Then what is the expected / intuitive behavior for that call? I think that I'd expect a `ConfigurationException` to be thrown rather than having `propValue` silently being set to `true`. Or maybe I shouldn't say that is the _intuitive_ behavior, but rather the _**safe**_ behavior. I think because copy/paste errors abound and it would be easy to make a typo with the property name, that's the only way that we're going to help people catch these bizarre accidental edge cases. Otherwise, we're going to cause them to curse at us because they'll have to figure up a debugger to figure out what the hell is going on in cases like this.

@xeno6696 and @jeremiahjstacey  - I'd like your opinion here. I know this requires yet another code change, but if we do a release and document it this way this (likely bug, IMO) will become a feature because changing it later on would break backward compatibility. If we were requiring JDK 9 or later we could just not export this method and keep it hidden, but since Java 8 is the minimal version, once we add this, it becomes available to the world.

src/main/java/org/owasp/esapi/configuration/EsapiPropertyManager.java
```diff
@@ -76,6 +76,18 @@ public Boolean getBooleanProp(String propertyName) throws ConfigurationException
         throw new ConfigurationException("Could not find property " + propertyName + " in configuration");
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Boolean getBooleanProp(String propertyName, Boolean defaultValue) {
+        for (AbstractPrioritizedPropertyLoader loader : loaders) {
+            return loader.getBooleanProp(propertyName, defaultValue);
```
kwwall, line 85: I think the way this is currently implemented, I don't think it will ever reach the final return on line 87 here. Maybe I'm wrong about that. Do we have a test for that case? Of course, if we change it as per my previous comment, we may still need it. Or may need to add a try / catch here and just rethrown any `ConfigurationException`.

src/main/java/org/owasp/esapi/configuration/StandardEsapiPropertyLoader.java
```diff
@@ -70,6 +70,24 @@ public Boolean getBooleanProp(String propertyName) throws ConfigurationException
...
+        if (property == null) {
+            return defaultValue;
+        }
+        if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
+            return true;
+        }
+        if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
+            return false;
+        }
+        return defaultValue;
```
kwwall, line 88: Depending on if we decide to change the behavior here in accordance to my first comment, we may wish to change line 88 to something along the following lines:
```java
    if ( property != null ) {
        throw new ConfigurationException("Property name '" + propertyName + "' has non-Boolean value of '" + property + "'.");
    }
```

src/main/java/org/owasp/esapi/configuration/XmlEsapiPropertyLoader.java
```diff
@@ -86,6 +86,24 @@ public Boolean getBooleanProp(String propertyName) throws ConfigurationException
...
+        if (property == null) {
+            return defaultValue;
+        }
+        if (property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes")) {
+            return true;
+        }
+        if (property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no")) {
+            return false;
+        }
+        return defaultValue;
```
kwwall, line 104: Ditto here. Similar to comment on **StandardEsapiPropertyLoader.java**.

src/main/java/org/owasp/esapi/reference/DefaultSecurityConfiguration.java
```diff
@@ -1441,21 +1441,47 @@ public Boolean getBooleanProp(String propertyName) throws ConfigurationException
...
+            if ( property == null ) {
+                return defaultValue;
+            }
+            if ( property.equalsIgnoreCase("true") || property.equalsIgnoreCase("yes") ) {
+                return true;
+            }
+            if ( property.equalsIgnoreCase("false") || property.equalsIgnoreCase("no") ) {
+                return false;
+            }
+            return defaultValue;
```
kwwall, line 1481: Ditto comment in **EsapiPropertyManager.java** about throwing `ConfigurationException` here.

src/test/java/org/owasp/esapi/configuration/EsapiPropertyManagerTest.java
```diff
@@ -304,6 +344,44 @@ public void testBooleanPropertyNotFoundByLoaderAndThrowException() {
...
+        } catch (IOException e) {
+            fail(e.getMessage());
+        }
+
+        boolean propertyValue = testPropertyManager.getBooleanProp(propertyKey, false);
+
+        // then
+        assertEquals(expectedPropertyValue, propertyValue);
+    }
+
```
kwwall, line 384: If we decide to revise this behavior to throw a `ConfigurationException` if `getBooleanProperty(String, Boolean)` is called and a non-Boolean property if found, we will need a new test here.

Ditto in **StandardEsapiPropertyLoaderTest.java** and **XmlEsapiProertperLoaderTest.java**.

kwwall (commented):
@mickeyz07 - I'm holding off on requesting changes until I get @xeno6696 and @jeremiahjstacey feedback about how getBooleanProp(String, Boolean) should treat a property that it encounters that has a non-Boolean value. I think we need to resolve that first. If we decide to do that, I will give you the option to finish it up, but if you don't, I'll probably approve this and then do those final changes myself in a separate PR. But getting this behavior of this new method is important and we'll likely only get one chance at getting it right.

@xeno6696 and @jeremiahjstacey -- waiting to hear back from you on this.

src/main/java/org/owasp/esapi/configuration/EsapiPropertyLoader.java
jeremiahjstacey: Overall, I think this is more scope than we should put time into to modify the opt-out request from the original issue.  I appreciate the initiative; however, we're adding this case to expand for something that could have been solved in two copies of an 6 line block.

This is overengineering in this scope.  We're blocking value, and a new possible release for this API change.

My suggestion is to take it out.  Use the duplicate snippet from my previous comment in the 2 factories and revert the changes in the DefaultSecurityConfiguration.  Create another task for the default API that we can address (reuse this code if you want), and complete it in a non-blocking scope.  We don't need this to fix the original problem, and as you can see the discussion of edge cases and caveats will continue to grow.  Let's get the opt-out behavior in in a first form, then improve upon it in another task to expand the SecurityConfiguration API to allow defaults to be specified.

kwwall: I concur. If in the future, you want to do a new PR to get a new interface for
```java
public Boolean getBooleanProp(String propertyName, Boolean defaultValue) throws ConfigurationException
```
I am fine with that, but as my review of this as surfaced, this is a little more nuanced than intuition alone might have us believe. If we are going to add that a future time, I think we need to start out with a GitHub issue and discuss the specifications there first, before waiting for an implementation to discuss them. So, to proceed with this PR, please rip that part of the code out and follow Jeremiah's suggestion.

src/main/java/org/owasp/esapi/logging/java/JavaLogFactory.java
```diff
@@ -80,7 +80,16 @@ public class JavaLogFactory implements LogFactory {
         boolean logApplicationName = ESAPI.securityConfiguration().getBooleanProp(LOG_APPLICATION_NAME);
         String appName = ESAPI.securityConfiguration().getStringProp(APPLICATION_NAME);
         boolean logServerIp = ESAPI.securityConfiguration().getBooleanProp(LOG_SERVER_IP);
-        boolean logPrefix = ESAPI.securityConfiguration().getBooleanProp(LOG_PREFIX, true);
+
+        boolean logPrefix = true;
+        try {
+            logPrefix = ESAPI.securityConfiguration().getBooleanProp(LOG_PREFIX);
+        } catch (ConfigurationException ex) {
+            System.out.println("ESAPI: Failed to read Log Prefix configuration. Defaulting to enabled" +
```
kwwall, line 88: Please change the output to something that explicitly mentions the property name on the output sent to stderr. E.g., something like this:
```java
 System.out.println("ESAPI: Failed to read Log Prefix configuration, " + LOG_PREFIX + ". Defaulting to enabled" +
```
That will make it easier for those getting this failure to know exactly what property to look for if they wish to change it in their **ESAPI.properties** file.

src/main/java/org/owasp/esapi/logging/slf4j/Slf4JLogFactory.java
```diff
@@ -70,7 +71,16 @@ public class Slf4JLogFactory implements LogFactory {
         boolean logApplicationName = ESAPI.securityConfiguration().getBooleanProp(LOG_APPLICATION_NAME);
         String appName = ESAPI.securityConfiguration().getStringProp(APPLICATION_NAME);
         boolean logServerIp = ESAPI.securityConfiguration().getBooleanProp(LOG_SERVER_IP);
-        boolean logPrefix = ESAPI.securityConfiguration().getBooleanProp(LOG_PREFIX, true);
+
+        boolean logPrefix = true;
+        try {
+            logPrefix = ESAPI.securityConfiguration().getBooleanProp(LOG_PREFIX);
+        } catch (ConfigurationException ex) {
+            System.out.println("ESAPI: Failed to read Log Prefix configuration. Defaulting to enabled" +
```
kwwall, line 79: Same as previous comment.

kwwall (changes requested):
Two lines where I'd like to see some very minor tweaks, otherwise LGTM.

kwwall (approved):
LGTM.

jeremiahjstacey (approved):
Looks good to me as well

kwwall:
I'm going to go ahead and bypass the 'branch protection' and do the 'squash and merge' because @mickeyz07 signed most of his commits and we reviewed all the code, so someone usurped his GitHub username and snuck something malicious past us that's one us.

Issue #851, opened 2024-08-11
Fix Typos


DarioViva42:
This is the first Issue requested in the following comment.
https://github.com/ESAPI/esapi-java-legacy/pull/850#issuecomment-2282343202

> One PR to only correct typos in public documentation (so .md files, comments in .properties / configuration files, and Javadoc.

> Every PR that you submit needs to have a corresponding GitHub issue created for it and the PR should reference that issue.

kwwall:
Closed via PR #852

PR #852, opened 2024-08-11
Fix Typos

DarioViva42:
Fix Issue https://github.com/ESAPI/esapi-java-legacy/issues/851

### DIFF
--- a/src/main/java/org/owasp/esapi/Encryptor.java
+++ b/src/main/java/org/owasp/esapi/Encryptor.java
@@ -162,8 +162,8 @@ public interface Encryptor {
      * </p>
      * @param ciphertext The {@code CipherText} object to be decrypted.
      * @return The {@code PlainText} object resulting from decrypting the specified
-     *            ciphertext. Note that the returned plaintext byte array
-     *            should be converted to a Java String using
+     *            ciphertext. Note that if it is desired to convert the returned
+     *            plaintext byte array to a Java String it should be done using
      *            {@code new String(byte[], "UTF-8");} rather than simply using
      *            {@code new String(byte[]);} which uses native encoding and may
      *            not be portable across hardware and/or OS platforms.
@@ -186,8 +186,8 @@ public interface Encryptor {
      * @param key        The {@code SecretKey} to use for encrypting the plaintext.
      * @param ciphertext The {@code CipherText} object to be decrypted.
      * @return The {@code PlainText} object resulting from decrypting the specified
-     *            ciphertext. Note that the returned plaintext byte array
-     *            should be converted to a Java String using
+     *            ciphertext. Note that if it is desired to convert the returned
+     *            plaintext byte array to a Java String it should be done using
      *            {@code new String(byte[], "UTF-8");} rather than simply using
      *            {@code new String(byte[]);} which uses native encoding and may
      *            not be portable across hardware and/or OS platforms.
--- a/src/main/java/org/owasp/esapi/Encryptor.java
+++ b/src/main/java/org/owasp/esapi/Encryptor.java
@@ -162,8 +162,8 @@ public interface Encryptor {
      * </p>
      * @param ciphertext The {@code CipherText} object to be decrypted.
      * @return The {@code PlainText} object resulting from decrypting the specified
-     *            ciphertext. Note that it it is desired to convert the returned
-     *            plaintext byte array to a Java String is should be done using
+     *            ciphertext. Note that the returned plaintext byte array
+     *            should be converted to a Java String using
      *            {@code new String(byte[], "UTF-8");} rather than simply using
      *            {@code new String(byte[]);} which uses native encoding and may
      *            not be portable across hardware and/or OS platforms.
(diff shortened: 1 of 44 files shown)


DarioViva42:
@kwwall I hope this is now manageable. Only 54 commits!

src/main/java/org/owasp/esapi/Encryptor.java
```diff
@@ -162,8 +162,8 @@ CipherText encrypt(SecretKey key, PlainText plaintext)
      * </p>
      * @param ciphertext The {@code CipherText} object to be decrypted.
      * @return The {@code PlainText} object resulting from decrypting the specified
-     *            ciphertext. Note that it it is desired to convert the returned
-     *            plaintext byte array to a Java String is should be done using
+     *            ciphertext. Note that the returned plaintext byte array
```
kwwall, line 165: This works, but I probably would have just fixed the actual typo here and changed the first 'it' to 'if' as was originally intended.

src/main/java/org/owasp/esapi/Encryptor.java
```diff
@@ -186,8 +186,8 @@ CipherText encrypt(SecretKey key, PlainText plaintext)
      * @param key        The {@code SecretKey} to use for encrypting the plaintext.
      * @param ciphertext The {@code CipherText} object to be decrypted.
      * @return The {@code PlainText} object resulting from decrypting the specified
-     *            ciphertext. Note that it it is desired to convert the returned
-     *            plaintext byte array to a Java String is should be done using
+     *            ciphertext. Note that the returned plaintext byte array
```
kwwall, line 189: Same as previous comment.

src/main/java/org/owasp/esapi/StringUtilities.java
```diff
@@ -90,7 +90,7 @@ public static boolean contains(StringBuilder input, char c) {
     }
 
     /**
-     * Returns the replace value if the value of test is null, "null", or ""
+     * Returns the replace value if the value of test is null, "null" (case-insensitive), or blank
```
kwwall, line 93: While you're changing this, please change 'replace' here to '{@code replace}'. I kept thinking it should be 'replaced' until I noticed it was the name of a formal parameter. Thanks.

src/main/java/org/owasp/esapi/User.java
```diff
@@ -380,7 +380,7 @@ public interface User extends Principal, Serializable {
     /**
      * Set the time of the last failed login for this user.
      *
-     * @param lastFailedLoginTime the date and time when the user just failed to login correctly.
+     * @param lastFailedLoginTime the date and time when the user just failed to log in correctly.
```
kwwall, line 383: No. Let's leave this as 'login'. The word login is even in the Cambridge dictionary and if people are searching, they likely would search for 'login' rather than 'log in'.

src/main/java/org/owasp/esapi/codecs/HTMLEntityCodec.java
```diff
@@ -198,7 +198,7 @@ private Integer parseNumber( PushbackSequence<Integer> input ) {
                 sb.appendCodePoint( c );
                 input.next();
 
-            // if character is a semi-colon, eat it and quit
+            // if character is a semicolon, eat it and quit
```
kwwall, line 201: Okay, I'll let this pass, but this sort of comment is not "public" in the sense that it's not going to end up in generated Javadoc. One of my reasons wanting to separate this type of comment from Javadoc comments is that we don't need to scrutinize these as much.

kwwall (changes requested):
LGTM; just a few places where I requested some minor changes.

src/main/java/org/owasp/esapi/codecs/HTMLEntityCodec.java
DarioViva42: Oh sh*t, I have cherry picked my way through all the commits and went through them again a second time, dropping all of them that did not match your criteria. This one commit must have slipped through. That was not my intention, and if you want, I can revert this commit again.

src/main/java/org/owasp/esapi/User.java
DarioViva42: Hey Kevin. https://dictionary.cambridge.org/dictionary/english-german/log-in?q=login
as far as I see it, cambridge also lists it as “log in” rather than “login”.
“login” is a noun, and “log in” is a verb.
https://www.reddit.com/r/webdev/comments/aml4hb/login_vs_log_in_are_we_using_these_wrong/
https://www.grammar.com/log_in_vs._login

Of course I can revert this commit, if you please, but I am still convinced that “log in” looks and sounds better when reading the sentence.

kwwall: How about we change it to "authenticate" then? Part of my reason is if people are searching through Javadoc for a term, I think they are unlikely to use "log in" or "sign on" because they are used to seeing them in login forms as a single word.

src/main/java/org/owasp/esapi/codecs/HTMLEntityCodec.java
kwwall: It's fine. There's only a few of those here and I reviewed them all. In your original PR there were a lot of these types, and I'm just saying as a rule, I really don't care that much if those "internal comments" are ever fixed unless they are flat out misleading (e.g., like they left out a 'not') so the logic of the comment is reversed or no one can really understand what the comment meant. But only people working on ESAPI code itself generally pay attention to that code. So, really low priority to fix these.

src/main/java/org/owasp/esapi/Encryptor.java
kwwall: Resolving this since your additional commit just revised this to 'if'.

src/main/java/org/owasp/esapi/StringUtilities.java
kwwall: Resolving this since it was addressed in commit [cbac2c8](https://github.com/ESAPI/esapi-java-legacy/commit/cbac2c8f9ee5a09dbdb09ec97d4fabb2642143cc).

kwwall (approved):
All good now.

PR #853, opened 2024-08-31
Improved documentation

DebajitKumarPhukan:
1. Updated version to latest available.
2. Added a quickstart maven example section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,10 @@ that's the version that will be used by default.
 <dependency>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
-    <version>2.5.4.0</version>
+    <version>2.5.3.0 [or later]</version>
     <classifier>jakarta</classifier>
 </dependency>
+<!-- Add the below dependency if the not using JakartaEE -->
 <dependency>
     <groupId>jakarta.servlet</groupId>
     <artifactId>jakarta.servlet-api</artifactId>
@@ -62,7 +63,7 @@ that's the version that will be used by default.
     <scope>provided</scope>
 </dependency>
 ```
-### Step 2: Download the ESAPI.properties and validation.properties from below locations and put it in your classpath.
+### Step 2: Create 2 properties file namely: ESAPI.properties and validation.properties and add them to your classpath. You can refer to below for boilerplate configurations.
 [ESAPI.properties](https://github.com/ESAPI/esapi-java-legacy/blob/develop/configuration/esapi/ESAPI.properties) <br/>
 [validation.properties](https://github.com/ESAPI/esapi-java-legacy/blob/develop/configuration/esapi/validation.properties)
 
@@ -78,6 +79,8 @@ public class Example {
    }
 }
 ```
+For more examples, please visit [https://owasp.org/www-project-enterprise-security-api/](https://owasp.org/www-project-enterprise-security-api/).
+
 # A word about ESAPI vulnerabilities
 A summary of all the vulnerabilities that we have written about in either the
 ESAPI Security Bulletins or in the GitHub Security Advisories may be found
--- a/README.md
+++ b/README.md
@@ -78,9 +78,6 @@ public class Example {
    }
 }
 ```
-
-
-
 # A word about ESAPI vulnerabilities
 A summary of all the vulnerabilities that we have written about in either the
 ESAPI Security Bulletins or in the GitHub Security Advisories may be found
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ fact, without the
 that's the version that will be used by default.
 
 # Quickstart - Maven Example
-#### Step 1: Add the required maven dependencies
+### Step 1: Add the required maven dependencies
 ```xml
 <dependency>
     <groupId>org.owasp.esapi</groupId>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ specific Jakarta version of ESAPI, in Maven, you would specify your ESAPI depend
 <dependency>
     <groupId>org.owasp.esapi</groupId>
     <artifactId>esapi</artifactId>
-    <version>2.5.3.0-SNAPSHOT</version>
```
kwwall, line 33: I don't want to have yet another place where we need to keep the ESAPI version current in the README.md file, so instead of 2.5.4.0, let's just drop the "-SNAPSHOT" and then somewhere add a note that says that the version has to be 2.5.3.0 _or later_. That way, we won't have to bother keeping that version up-to-date.

README.md
```diff
@@ -46,6 +46,38 @@ fact, without the
 ```
 that's the version that will be used by default.
 
+# Quickstart - Maven Example
```
kwwall, line 49: First off, I guess I should have really pointed at our "[Using ESAPI with Jakarta EE Servlet API Specification 5.0 and later](https://github.com/ESAPI/esapi-java-legacy/wiki/Using-ESAPI-with-Jakarta-EE-Servlet-API-Specification-5.0-and-later)" wiki page instead. Pretty much all that you need to know is over there about this is there, and if it isn't, that's where it should be placed, not here. (Unfortunately, only the core team can make PRs for that; it seems to be a GitHub thing.) I just checked, and our current README file is already 8 pages. If anything, I'd like to make it shorter and more some of the nonessential content to our wiki page. The only reason that I added a section for "Jakarta EE Support" is because not too long ago I was getting slammed with personal emails asking me "does ESAPI support Jakarta EE yet". Adding that cut down on that, but I definitely should have linked to the corresponding wiki page from here.

So, look at the wiki page that I've referenced and if you think that the wiki page needs updated, please create a GitHub issue outlining the suggested change and ideally include a "patch" file that I can apply to update it and we can handle it that way (since you can accept a PR against the wiki pages.)

README.md
```diff
@@ -46,6 +46,38 @@ fact, without the
...
+### Step 1: Add the required maven dependencies.
+```xml
+<dependency>
+    <groupId>org.owasp.esapi</groupId>
+    <artifactId>esapi</artifactId>
+    <version>2.5.4.0</version>
+    <classifier>jakarta</classifier>
+</dependency>
+<dependency>
```
kwwall, line 58: I'm not sure I would show this dependency. For the older Java EE Servlet API, we have that as:
```xml
        <dependency>
            <groupId>javax.servlet</groupId>
            <artifactId>javax.servlet-api</artifactId>
            <version>3.1.0</version>
            <scope>provided</scope>
        </dependency>
```

Usually,, if you are using Jakarta EE, the Jakarta Servlet API is either pulled in by the JDK itself (I think for versions later than Java 17 that's the case, but I may be wrong about the version number) or your pulling it in through another dependency like Spring Boot 3.x, Spring Framework 6.x, or Tomcat 10.x or later.

The problem with including this dependency is that someone would then expect us to keep this updated to the latest version and since we just convert the byte code for the Jakarta version of ESAPI, we really don't explicitly check compatibility with any particular version of the Jakarta EE Servlet Spec.

README.md
```diff
@@ -46,6 +46,38 @@ fact, without the
...
+    <artifactId>jakarta.servlet-api</artifactId>
+    <version>6.1.0</version>
+    <scope>provided</scope>
+</dependency>
+```
+### Step 2: Download the ESAPI.properties and validation.properties from below locations and put it in your classpath.
+[ESAPI.properties](https://github.com/ESAPI/esapi-java-legacy/blob/develop/configuration/esapi/ESAPI.properties) <br/>
+[validation.properties](https://github.com/ESAPI/esapi-java-legacy/blob/develop/configuration/esapi/validation.properties)
+
+### Step 3: Let's say, you want to remediate log injection vulnerabilities. Below is an example to achieve it.
```
kwwall, line 69: Adding examples in the README file I think is a bad precedent to set. As an alternative, we could possibly add this to the GitHub wiki or create an 'examples' subdirectory under "documentation" and add stuff there.

kwwall (changes requested):
See my other comments for the changes requested. IMO, our README is already too long.

kwwall:
@DebajitKumarPhukan - I'd like to do a release soon (no later than 9/22). Do you think you can make the requested changes by then? If not, these changes will have to wait until the end of November.

README.md
kwwall: @DebajitKumarPhukan - I have a bunch of changes (mostly reorganization and simplification) to our README.md, but I was waiting until you finished this PR so we wouldn't get merge conflicts. I may still move that content to our wiki page, or at least change it and drop it into a collapsible section. But our README is already just a shade over 8 pages if you were to print it. I think that's about 3 or 4 pages too long. But this is good enough for now.

README.md
kwwall: I'll have to think about this. I don't want or team to have the burden to have to keep dependency versions in the **README.md** in sync with that is reflected in our **pom.xml.**. This may be content I move to our a local GH wiki page. I'd been enough of a PITA just for me to remember to update _ESAPI_ versions in the README. And I can't recall anyone who is using ESAPI with only Java SE. Probably 80% of the ESAPI components require the Servlet API from JavaEE (or JarkataEE, depending on your perspective). And those who start out using ESAPI from the start is likely a minority. Based on both secure code reviews that I've done and feedback from developers from private emails and SO posts, most of them only start using ESAPI when some SAST tool identifies vulnerabilities and suggests ESAPI as a potential solution, and usually in the source-to-sink taint path the follow, the 'source' is most often something like an HTTP parameter associated with the Servlet API.

kwwall (approved):
I'm fine with the changes, as-is, although I likely will move some of this to our wiki page. These additions are easily going to push our README.md to 9 pages.

kwwall:
Since this is only documentation changes, I am going to merge PR this without the approval of @xeno6696 or @jeremiahjstacey.

PR #869, opened 2025-01-14
use latest stable release of commons-collections4 (fixes #868)

### LINKED ISSUE #868: Do not depend on commons-collections4 milestone (use 4.4 instead)
#### Describe the bug
esapi uses a milestone release of commons-collections4. These are for testing only. Using them forces downstream projects to depend on it as well.

2.6.0.0

reschke:


kwwall:
Rejecting this PR for reasons that I and @xeno6696 stated in (now closed) issue #868. We can re-open if you want to do a PR that eliminates this Collections 4 direct dependency for us, but that likely will require more complete JUnit tests than currently exist for the core team to get a comfort level that we are not unintentionally breaking something.

PR #871, opened 2025-02-03
Bump to org.apache.commons:commons-configuration2

picsouds:
Bump from commons-configuration:commons-configuration:1.10 to org.apache.commons:commons-configuration2 (fixes many CVEs)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <version.jmh>1.37</version.jmh>
-        <version.findsecbugs>2.0.0-M3</version.findsecbugs>
+        <version.findsecbugs>1.13.0</version.findsecbugs>
         <version.fluido>2.0.0-M11</version.fluido>  <!-- 2.0.0 fails on 'mvn site', but this works. -->
         <version.powermock>2.0.9</version.powermock>
         <version.spotbugs>4.8.6</version.spotbugs>
@@ -194,9 +194,9 @@
                  -->
         </dependency>
         <dependency>
-            <groupId>commons-configuration</groupId>
-            <artifactId>commons-configuration</artifactId>
-            <version>1.10</version>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-configuration2</artifactId>
+            <version>2.11.0</version>
             <exclusions>
                 <!-- excluded because multiple dependencies import newer version. -->
                 <exclusion>
@@ -424,6 +424,8 @@
                      </configuration>
                 </plugin>
             </plugins>
+
+
         </pluginManagement>
 
         <plugins>
@@ -706,6 +708,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${version.surefire}</version>
+                <configuration>
+                    <argLine>-Duser.language=en</argLine>
+                </configuration>
             </plugin>
 
             <plugin>
--- a/src/main/java/org/owasp/esapi/reference/accesscontrol/policyloader/ACRParameterLoader.java
+++ b/src/main/java/org/owasp/esapi/reference/accesscontrol/policyloader/ACRParameterLoader.java
@@ -1,6 +1,6 @@
 package org.owasp.esapi.reference.accesscontrol.policyloader;
 
-import org.apache.commons.configuration.XMLConfiguration;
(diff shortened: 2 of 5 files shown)


kwwall:
> Bump from commons-configuration:commons-configuration:1.10 to org.apache.commons:commons-configuration2 (fixes many CVEs)

@picsouds - Let's start with "what specific CVEs do you (or @smmsit) think are being fixed here? As I mentioned in GH issue #872, I don't see any legitimate ones. And I don't think just because a library is old is not reason enough to make changes.

My second major objection--and one that I thought I made clear--is that has an inexcusable set of Junit tests related to the ESAPI-AccessControlPolicy.xml policy file. Therefore, because of the scarcity of tests, it is not surprising why this PR "maintains compatibility especially for ACRParameterLoaderHelper". The problem is, without additional testing, we could accidentally breaking something for one of the rare cases where an ESAPI client is actually using AccessController. Therefore, if we are going to merge this PR, I for one (and I assume that @xeno6696 and @jeremiahjstacey are in this camp as well), we are going to need some additional tests written to ensure we haven't accidentally introduced an incompatibility that's going to break someone's code. So given that and that I see no compelling evidence that this PR addresses in terms of known vulnerabilities, I am more inclined to just reject this PR. So, if you want to get this addressed, then consider writing additional tests cases for it. Thanks.

picsouds:
Hi @kwwall
I'm totally agree with you and the word "many" is inappropriate (2 CVEs false positive) in my PR (I apologize for that).
But I think that in terms of modernization, it could not be bad to upgrade this dependency (although as you are previously mentioned no end of life mentioned on apache-configuration).
OK to add more additional testing (with signed commit - didn't read the doc carefully) to cover more cases.

kwwall:
> Hi @kwwall I'm totally agree with you and the word "many" is inappropriate (2 CVEs false positive) in my PR (I apologize for that). But I think that in terms of modernization, it could not be bad to upgrade this dependency (although as you are previously mentioned no end of life mentioned on apache-configuration). OK to add more additional testing (with signed commit - didn't read the doc carefully) to cover more cases. Thank you very much for this exchange. 👍

Again, backwards compatibility is very important. If your PR breaks c ESAPI client's code, the core team is the one that will hear about it, not you. So, if I am going to merge this, I want to see a bunch of extensive tests implemented and in place _first_, that tests it as implemented with Common Configuration 1.10. That way we will know exactly what exceptions 1.10 throws for any edge cases. Without that, we won't know (with extensive code review or Javadoc review) if Commons Configuration 2.x changes any exceptions that could affect client code. (I have caught things in code reviews of proprietary libraries where a colleague changed a FileNotFoundException that was previously thrown to instead throw an IOException since there were some new edge cases where exceptions thrown. I remarked that could break backward compatibility of existing library client code if the client code were catching 'FileNotFoundException' rather than some base class. So, yeah; it can happen.)

The other thing is, you said there were 2 false positives identified. I want to know the specific CVE IDs that you believe are true positives that your SCA tool has identified. Because I see none. And even if a couple of them are valid, that hardly means there is an exploitable path in the manner that ESAPI uses Commons Configuration 1.10. (For example, if the CVE was associated with some Commons Configuration class that is not even used by ESAPI, it would be exploitable. I will gladly do the research, and write it up as a Security Bulletin. Because even if it is not exploitable and we decide to fix upgrade to Commons Configuration 2.11 anyhow, ESAPI users will want to know.

@xeno6696 or @jeremiahjstacey - Do either of you have anything to add?

picsouds:
like you said in https://github.com/ESAPI/esapi-java-legacy/issues/872, CVE-2024-29131 and CVE-2024-29133 (these are both false positives, as these both are only applicable to the 2.x releases of Commons Configuration - see https://github.com/dependency-check/DependencyCheck/issues/6704).
You can close this PR, it requires few additional tests.

kwwall:
My question is, and always has been partly about whether you believe
there's any true positive CVEs. If yes, please state them, and I'll
consider the options. If they are exploitable, we ought to do a PR for
this. It's just that I have seen no evidence of any true positives of CVE
IDs any SCA tool has found. However, if there are some and they are
exploitable via ESAPI,  the ESAPI core team will help with the new JUnit
tests.

picsouds:
For me, there is no positive CVE on commons-configuration:commons-configuration:1.10 identified to date.
I'm closing this PR which is not an update of commons-configuration:commons-configuration to fix CVEs but rather to move to org.apache.commons:commons-configuration2 (but this may introduce other problems - breaking changes, etc...).

PR #879, opened 2025-05-18
2.6.1.0 release prep

kwwall:
- Updated AntiSamy from release 1.7.7 to 1.7.8 which addresses the potentially exploitable vulnerability CVE-2025-27820. There is  slim possibility that this could affect ESAPI users who have allowed certain CSS mark-up constructs to the AntiSamy policy file that they are using. However the default ESAPI AntiSamy policy file (**antisamy-esapi.xml**) does not permit CSS mark-up of any sort out unless it has been modified by the ESAPI client.
- Other minor updates to pom.xml
- Added new release notes for release 2.6.1.0


xeno6696 (approved):
Straightforward, I'll miss the dev stats but I get it.

PR #882, opened 2025-06-02
Fix guessed release date for 2.6.1.0 to its actual release date.

kwwall:
This was to fix up things to reflect reality and was needed since Maven Central took a bit longer than anticipated the last time.

@xeno6696 / @jeremiahjstacey - Note that there will shortly be a 2nd PR following this on a 2.6.2.0 branch as per the usual process.


xeno6696 (approved):
Trivial.  No issues.

PR #886, opened 2025-06-09
Javadoc enhancements

kwwall:
Work on fixing up some Javadoc and tweak test version of ESAPI.properties.


kwwall:
@xeno6696 @jeremiahjstacey @sempf @hackermatic - The Javadoc is ready to review. No code changes to this PR.

src/test/resources/esapi/ESAPI.properties
```diff
@@ -578,3 +578,37 @@ Validator.AcceptLenientDates=false
...
+#                       method to indicate here when the method name alone,
+#                       absent from its signature, is ambiguous, so it is
+#                       best to avoid those if at all possible!
+#
+#                       An example of that would be something like:
+#                           org.owasp.esapi.reference.DefaultValidator.getValidPrintable
+#                       which has 4 interfaces so currently, there's no way to
+#                       specify a specific one.
+#       
+ESAPI.enableLegCannonModeAndGetMyAssFired.methodNames=org.owasp.esapi.reference.DefaultEncoder.encodeForSQL
```
jeremiahjstacey, line 608: I agree with the intent, but we may want to adjust the property name before releasing to the public.  I expect folks will be using this as a stop-gap to keep code from breaking as they're working through migration processes.

src/test/resources/esapi/ESAPI.properties
```diff
@@ -578,3 +578,37 @@ Validator.AcceptLenientDates=false
...
+
+# Normally you would put some text here (that will be logged) that provides some
+# justification as to why you have enabled these functions. This can be
+# anythuing such as a Jira or ServiceNow ticket number, a security exception
+# reference, etc. If it is left empty, it will just like "Justification: none".`
+ESAPI.enableLegCannonModeAndGetMyAssFired.justification=blah,blah. Please don't fire my @$$. Ticket # 12345
```
jeremiahjstacey, line 614: I would prefer to not allow this feature for client-specific messages in library workflows.
We should have a message that is either attributed to an Exception or logged at a Warning level which states the feature that is bound for deprecation and the date which it will be removed (per deprecation policy).

Clients can track this however they want in their own environments.

EG:
```
encodeForSQL(Codec codec, String input) {
  String deprecationMsg="ESAPI.DEPRECATION-6//9/2026   ESAPI.encodeForSQL method use should be replaced with ....blah blah blah";
if ( ! deprecationsEnabled("encodeForSQL")) {
  throw new NotConfiguredByDefaultException(deprecationMsg);
} else {
  log.warning(deprecationMsg);
}
```

src/test/resources/esapi/ESAPI.properties
kwwall: It caught your attention didn't it?

src/test/resources/esapi/ESAPI.properties
kwwall: We will still have our own messages and this might not want to limit it to just things that we've deprecated. But the intent here is that many companies require approval of use of dangerous and/or deprecated methods and this would allow them to provide an identifying security exception ID that InfoSec could cross reference in audit logs. I would likely log it independently rather than make it part of the exception message (which you never know, may end up being displayed to the end user, which might have been your concern). But the ESAPI's safe logging should be able to handle whatever get's thrown at it, otherwise it's not really "safe".

Anyhow, that's the thought behind it. Mostly to help track in audit logging, which was @xeno6696's idea. I don;t think I can compose any sort of structure here as different companies use different things to ID long-standing security exceptions.

src/main/java/org/owasp/esapi/Encoder.java
sempf, line 1: Line 160, advice not advise

src/main/java/org/owasp/esapi/codecs/DB2Codec.java
sempf, line 1: Line 24 finding not find

src/main/java/org/owasp/esapi/codecs/MySQLCodec.java
sempf, line 1: Line 23 chances not changces

kwwall: Corrected via 2d444b903d59a851fb0805f17f04342deecf9269

src/main/java/org/owasp/esapi/codecs/OracleCodec.java
```diff
@@ -18,10 +18,14 @@
...
 /**
- * Implementation of the Codec interface for Oracle strings. This function will only protect you from SQLi in the case of user data
- * bring placed within an Oracle quoted string such as:
- *
- * select * from table where user_name='  USERDATA    ';
+ * Implementation of the {@link org.owasp.esapi.codecs.Codec} interface for Oracle DB strings. 
+ * This function will only protect you from SQLi in limited situations.
+ * To improve your changces of success, you made also need to do some
+ * additional canonicalization and input validation first. Before using this class,
+ * pleaes be sure to read the "SECURITY WARNING" in
```
xeno6696: please

src/test/resources/esapi/ESAPI.properties
xeno6696: I think for log statements, including the CVE number directly in the logs would be proper.  Not sure we should push that however until we're ready to be public-public.

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -147,10 +147,28 @@
...
+ * over the years have made every effort to be cautious--often going to extremes
+ * to make "safe harbor" decisions on harmful inputs other similar encoders assume are already safe
+ * (we did this to in order to protect the client's users from buggy browsers that don't adhere
+ * to the W3C HTML specications)&em;the various {@code Codec} implemtations can offer
+ * NO GUARANTEE of safety of the content being encoded or decoded. Therefore,
+ * it is highly advised to practice a security-in-depth approach for everything you do.
+ * By following that advise, you will minimize the impact and/or likelihood of any
+ * vulnerabilities from bugs in the ESAPI code or accidental misuse of the ESAPI
+ * library on your part. In particular, whenever there are cases where cients use
+ * any of these {@link org.owasp.esapi.codecs.Codec} classes drectly, it is highly
```
xeno6696, line 163: directly

src/main/java/org/owasp/esapi/Encoder.java
```diff
@@ -395,25 +413,69 @@ public interface Encoder {
...
+     * The this method attempts to ensure make sure any single-quotes are double-quoted
+     * (i.e., as '', not double-quotes, as in &quot;). Another possible approach
+     * is to use the {escape} syntax described in the JDBC specification in section 1.5.6.
      * However, this syntax does not work with all drivers, and requires
      * modification of all queries.
-     *
+     * </p><p>
+     * <b>SECURITY WARNING:</b> This method is <u>NOT</u> recommended. The use of the {@code PreparedStatement}
+     * interface is the preferred approach. However, if for some reason
+     * this is impossible, then this method is provided as significantly weaker
```
xeno6696, line 428: "as a significantly..."

src/main/java/org/owasp/esapi/codecs/MySQLCodec.java
```diff
@@ -19,29 +19,36 @@
 
 /**
  * Codec implementation which can be used to escape string literals in MySQL.
- * </br>
- * Implementation accepts 2 Modes as identified by the OWASP Recommended
- * escaping strategies:
+ * This function will only protect you from SQLi in limited situations.
+ * To improve your changces of success, you made also need to do some
```
xeno6696: chances mispelled

xeno6696:
I approve, just in the dentists chair and don’t have time to find the right button

PR #887, opened 2025-06-26
fix: CVE-2024-47554 vulnerability

gustavonj:
The following vulnerability is fixed with an upgrade:
-https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-47554

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -221,7 +221,7 @@
         <dependency>
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
-            <version>1.5</version>
+            <version>1.6.0</version>
             <exclusions>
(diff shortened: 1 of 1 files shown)


paulbors:
Can we get this in please?

paulbors:
We can wait 4 days :)

Beats having to push a custom artifact into our organization's artifact mirror

kwwall:
Usually it easy to just explicitly exclude the vulnerable dependency
version in the application and the include the patched version directly.  I
can assure you in the case, it won't break ESAPI. I ran our JUnit tests
with the patched version and all tests passed.

paulbors:
That's the thing... one must have that guarantee before starting to mess with transient dependencies.

You just gave us the green light for this one artifact.

PR #901, opened 2025-11-01
More tests

jeremiahjstacey:
I got bored and thought I'd pick out some low-hanging classes for a couple more tests.

### DIFF
--- a/src/main/java/org/owasp/esapi/waf/actions/DoNothingAction.java
+++ b/src/main/java/org/owasp/esapi/waf/actions/DoNothingAction.java
@@ -22,10 +22,6 @@ package org.owasp.esapi.waf.actions;
  */
 public class DoNothingAction extends Action {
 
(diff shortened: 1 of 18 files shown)


src/main/java/org/owasp/esapi/waf/actions/DoNothingAction.java
```diff
@@ -22,10 +22,6 @@
  */
 public class DoNothingAction extends Action {
 
-    public boolean failedRule() {
-        return this.failed;
-    }
```
jeremiahjstacey, line 27: This method duplicates the super class' implementation.  No point in it.

jeremiahjstacey:
I think this bumps coverage ~.2%

xeno6696 (approved):
I honestly didn't realize I'd been sitting on this PR for this long.  Apolgies, this year hasn't been my best.

Issue #906, opened 2026-02-05
SecurityWrapperResponse.setHeader does not handle nulls


ThrawnCA:
#### Describe the bug

Passing `null` to either the name or value parameters of `SecurityWrapperResponse.setHeader` will result in a `NullPointerException` (within `StringUtilities.stripControls`). This is almost always undesired behaviour, and contrary to the [latest Servlet specification](https://github.com/jakartaee/servlet/blob/main/api/src/main/java/jakarta/servlet/http/HttpServletResponse.java), which states how null values should behave (null header name is a no-op, null header value is a deletion).

#### Specify what ESAPI version(s) you are experiencing this bug in

2.7.0.0

#### To Reproduce

1. Create an application that wraps the request and response in `SecurityRequestWrapper`.
2. Attempt to delete a header using `setHeader("myHeader", null)`
3. `NullPointerException`

#### Expected behavior

Null values should be ignored by ESAPI and gracefully passed to regular code.

PR #907, opened 2026-02-05
gracefully handle null within 'stripControls', #906

ThrawnCA:


jeremiahjstacey:
@ThrawnCA 
It would be greatly appreciated if you could also add a test in the StringUtilitiesTest for strip controls.
There appear to be no tests for this method in that class currently.

Bare minimum would be a simple validation that when null is provided, null is returned.  If you're willing and able to provide more case coverage that would certainly be beneficial, but beyond the scope of this change.

Thank you for the contributions to the project.

ThrawnCA:
> @ThrawnCA It would be greatly appreciated if you could also add a test in the StringUtilitiesTest for strip controls. There appear to be no tests for this method in that class currently.

Done.

I notice that this project seems unready for Java 17. It fails module access checks. Also, several tests within `EncryptedUtilsPropertiesTest` fail because `Properties.store` attempts to call `ReferenceEncryptedProperties.entrySet`, which is specifically disallowed.

FYI In order to make the Surefire plugin run successfully (apart from EncryptedPropertiesUtilsTest), I updated its configuration locally to:

    <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>${version.surefire}</version>
        <configuration>
            <argLine>
                --add-opens java.base/java.lang=ALL-UNNAMED
                --add-opens java.base/java.util=ALL-UNNAMED
                --add-opens java.base/java.util.regex=ALL-UNNAMED
                --add-opens java.base/java.text=ALL-UNNAMED
                --add-opens java.logging/java.util.logging=ALL-UNNAMED
            </argLine>
        </configuration>
    </plugin>

jeremiahjstacey (approved):
Looks good to me.

